PR #5, opened 2024-01-22
Fixes For SafeStart

Njpatman:
- SafeStart Time is now sync'd to the server

- EH is now checked and re-applied periodically to make it compatible with Reforger Lobby

- Removed cap on players speaking range.

- Had the EHs apply to all players, so we removed any possibility of people accidentally shooting each other.


Njpatman:
This is ready to be pushed through in its current state.

tanaaKa:
Will review

scripts/Game/Systems/COA_SafeStart.c
```diff
@@ -193,95 +184,107 @@ class CRF_TNK_SafestartComponent: SCR_BaseGameModeComponent
...
+		if (status) { // Turn on safestart
+			if (m_SafeStartEnabled) return;
 			
 			m_iTimeSafeStartBegan = GetGame().GetWorld().GetWorldTime();
 			m_SafeStartEnabled = true;
 			m_iSafeStartTimeRemaining = 35;
+			
 			GetGame().GetCallqueue().CallLater(CheckStartCountDown, 5000, true);
-			ToggleSafeStartEHs();
+			GetGame().GetCallqueue().CallLater(UpdateServerWorldTime, 250, true);
```
tanaaKa, line 208: Does this really need need to run every 250ms?

Njpatman: I'd prefer it that way to make the time display as smooth as possible, but it could run at 500 or 1000ms

tanaaKa: Yeah I mean, so far, enfusion has proved to be efficient, but it's good to ask the question. It's your call.

Njpatman: I like how it is right now; I will push it through.

PR #15, opened 2024-03-28
Add Container Craze TvT Attack/Defense Mission

Pi123263:
For real this time, am a smart IT big brain man


Pi123263:
rel.: https://github.com/CoalitionArma/Coalition-Reforger-Framework/pull/13

tanaaKa:
I dont see anything wrong with it, but you'll want to add a mission config file. In that mission description, make sure to add a ratio of some sort.

tanaaKa:
Also the defending soviets will have a massive advantage here. How does the m113 aid blufor? If anything, maybe blufor gets an aerial vehicle. Your call here.

Pi123263:
Ah yep adding the description
The M113 is throught of as a long range suppression and semi-AT weapon in case the soviets decide to use any of the vehicles (especially BTR guns), which the 50 cal should be able to pen from a good bit out afaik?

tanaaKa:
Multiple BTRs will outclass the m113 tho. There's four that the defenders WILL use, making any assault by them irrelevant.

Pi123263:
Yeah thats fair, I rebalanced them a bit 
-removed the turrets from 2 of them (so they're pretty much just props now)
-removed the fuel from the 2 with guns so they static 
-added a few more containers to limit the fields of fire of the guns

PR #16, opened 2024-03-29
Add Silo Security Misison

Pi123263:
Small TvT mission with infantry focus centered around the Rocket Silo on Gogland


tanaaKa:
This silo is wild. I didn't even know it was on the map. 

- If this is meant to be Attacker 3:2 Defender, give the attacker a couple more squads just in case.
- While your at it, add an objective entity for both sides (I forgot to teach you this Wednesday)
image
Read the component fields and it should be fairly straight forward. Adding this objective entity enables the AAR screen so it's not blank. **You should probably do this for container craze as well and just include it in this PR.**
- Make sure the slotting screen is in order. It's based on the order defined in the hierarchy list.
- Experiment with the callsigns in the faction manager. Use gambit or ridge raid as an example if needed. Here is what it looks like:
image

Pi123263:
Yeah the silo is really cool, been wanting to use it in a TvT for a while
I really hope the defenders dont just get slaughtered outside and actually use it
gonna commit the suggestions in a few mins

Pi123263:
Also I put BLU 3:1.5 OPF (blu def, opf atk) in for the initial commit to try and balance it, I'm open for suggestions on that though

tanaaKa:
> Also I put BLU 3:1.5 OPF (blu def, opf atk) in for the initial commit to try and balance it, I'm open for suggestions on that though

Okay, mission config states BLU 3:2 OPF, so that'll need fixed too.

> Yeah the silo is really cool, been wanting to use it in a TvT for a while I really hope the defenders dont just get slaughtered outside and actually use it gonna commit the suggestions in a few mins

If you're worried about that, explicitly state what you want from them in a description file.

Pi123263:
Commits out

PR #18, opened 2024-03-29
Add Opdrew

MrPvTDagger:
4:1 Ratio

Blufor is a saboteur group trying to complete as many objectives while minimizing contact with opfor, blufor has night vision and thermals while opfor only has flashlights. Visibility should be really bad so should allow blufor to sneak around. Still deciding if I want opfor to have respawns as well. Opfor has a heli for QRF as well but that up to how the PL uses it

Win conditions are:
Blufor team is wiped = Opfor win
Blufor complete at least three main objectives, collects one intel and gets to the extraction point = Blufor win (They just need to throw a m67 or GP on the location of the objectives to destroy them It will give notifications when they are destroyed (Replication of this needs testing)

Rules:
no picking up NVG or radios
if respawns are on try to follow "New Life Rule" but I would think they will get updates on enemy positions quickly anyways after fire fights start.

I need to add this info mission description and etc to the lobby menu still.

This bit of a unique mission so hopefully it is balanced and goes well


tanaaKa:
Feedback:
- The triggers should check for the ammo dump being destroyed. Currently they check for a grenade entering the trigger which is a bit hacky. Use `OnFrame`
image
Misspelled 
image
**Wait**.... is this on _purpose_?

- You need to add objective entities. I forgot this yesterday. Adding the OBJ entities enables the debrief screen. Reference https://github.com/CoalitionArma/Coalition-Reforger-Framework/pull/16#issuecomment-2027425801 for instruction.

> if respawns are on try to follow "New Life Rule" but I would think they will get updates on enemy positions quickly anyways after fire fights start.

We don't plan on respawns, and truth be told, you should be stating whether or not this mission supports them.

OPFOR needs more thought out objectives. Right now they just feel like set dressing for BLUFOR.

MrPvTDagger:
- Yeah it's definitely hacky, I thought it was the best way to do it as they are map objects so I don't believe you can actually destroy them so nothing to check for. Might be worth switching them to ammo trucks and fuel trucks, Can you destroy ammo crates? For the power station & radio tower I still think best to use the trigger that detects the explosive. Might spawn a fire entity to give a visual indication rather than the pop up notification. 
- Do we have C4 in the mod pack? Use that instead of the grenade.
- English hard
- Will add objectives
- Didn't explain that part very well, What I meant to say was that respawns are enabled with waves for opfor only but I was undecided if I wanted to use them or not just from a gameplay stand point was more of a note to myself.
- When you say dressed to blufor do you mean it will be too easy for blufor to win vs opfor? I feel like it should be like that due to the ratio being so different. Or is it you don't feel like opfor has enough to do?

Njpatman (changes requested):
Hi Dagger, please update this PR to the latest version of the CRF or close it out.

MrPvTDagger:
Should be good now.

Njpatman (dismissed):

Njpatman:
Awaiting final push from Tanaka

tanaaKa:
Need the .rdb file conflict resolved

Njpatman (approved):

Njpatman:
Awaiting final push from Tanaka

PR #21, opened 2024-03-30
Feature: Add more prefab groups to be used by mission makers

Pi123263:
This PR will:
-Move the existing groups into a legacy subfolder while not changing them to remain backwards compatible
-Add new groups for both US and USSR forces organized into subfolders
--Command
--Infantry
--Vehicles
--Support
--Other (Currently empty, exists for special group prefabs etc.)


Njpatman:
Add a civ/insurgent faction pls

PR #22, opened 2024-03-30
Fixes #19

### LINKED ISSUE #19: Modern Russian Gear Script no Flashlight and Radio
There is an issue with Russian modern loadouts, They are not getting the flashlight and radio even though it in their inv component.
image
image

MrPvTDagger:
As @tanaaKa mentioned in #19 there are no slots on the vest cause they are on the jacket... (Looking at you RHS Team)

- Moved the range finder to the backpack as no room in the jacket for it hence it was being removed.
- Switched the RU radio to 148 as it was 187 which is a backpack radio.


tanaaKa:
We dont want to use the RHS 148. It's buggy. Give them the default reforger soviet radio and we can close this.

MrPvTDagger:
It should already be the default reforger soviet radio they are both called 148.

MrPvTDagger:
The RHS one is 187

tanaaKa:
Well that's dumb

PR #43, opened 2024-04-22
Create a Search & Destroy variant of Container Craze called 'Container Search'

Pi123263:
This PR will:
-Add a new Version of Container Craze called 'Container Search' using the Search & Destroy gamemode framework


Njpatman (approved):
Looks good, awaiting push by Tanaka

PR #44, opened 2024-04-23
Update

Arkangel-Vigilamus:
Now on a less archaic version of CRF!
(I think the kill trigger is broke after the revert, might need a look)


Njpatman (dismissed):
Looks good, fix that conflict tho

Worlds/Lobby/Eden/COA Rally 2_Layers/default.layer
```diff
@@ -0,0 +1,327 @@
...
+  angleZ -0.562
+ }
+ S105_rally89 {
+  coords 1399.047 47.314 6016.998
+  angleX 2.305
+  angleY -172.383
+  angleZ 3.013
+ }
+}
+>>>>>>> Stashed changes
```
tanaaKa: There's a bunch of git comments in these files that shouldn't be there.

tanaaKa (changes requested):

Worlds/Lobby/Eden/COA Rally 2_Layers/default.layer
tanaaKa: I dont have an elegant solution to remove these

PR #45, opened 2024-04-23
Rally Meaux Update 

Arkangel-Vigilamus:
Removed git comments(?)


Njpatman (approved):
It looks good. Remember to force ignore changes in the resourcedatabase.rdb from here on out.

PR #47, opened 2024-04-24
Naming Convention Standardization

Njpatman:


Njpatman:
We will need this out as soon as possible so that any missions/scripts in the works can be updated without too much headache.

I did this because if I hadn't, we would have built up so many systems that do COA and CRF interchangeably, and I just can't have that.

tanaaKa (approved):

Pi123263:
Might be good to add a small entry to the readme or wiki about the naming convention for people to look at/to send to people

Njpatman:
Nah, the biggest thing is keeping everything up to date with the CRF_ prefix just so we don't clash with any potential future mods.

PR #49, opened 2024-04-24
Fix player amounts, squad names, assets, squad arrangement and other ambient things

Pi123263:
See title, PR includes various fixes for the 'Container Collection' of missions


Njpatman:
pls update the .conf file to the new player count

Pi123263:
Should both be updated? Or did I miss some?

Njpatman:
Also, have you verified that S&D is fully working in your version? When I went through the naming convention stuff, I had to change the name of a trigger from asitetrigger2 to bsitetrigger.

Njpatman:
> Should both be updated? Or did I miss some?

image

Both aren't correct in their file names

Pi123263:
Ah yeah forgot the filenames
As for the site triggers, I just checked and I've got an a and b site trigger, however it actually doesnt seem to be working right? Is there another place I have to rename them?
It spawns both generator objects, but it seems like both of them are pointing to the A site script wise
image

Njpatman:
> Ah yeah forgot the filenames As for the site triggers, I just checked and I've got an a and b site trigger, however it actually doesnt seem to be working right? Is there another place I have to rename them? It spawns both generator objects, but it seems like both of them are pointing to the A site script wise image

It uses the rotation of the trigger, so rotate the triggers however you want the rotation of the generator to be

Pi123263:
Sorry I didnt mean pointing as in rotation, but rather that both triggers seem to be trying to 'access' the A site
Through some more testing just now I found out that it works fine if the bSite is blown up first and then the aSite second, however if the aSite is blown up first and the bSite second the following error pops up when trying to plant the bomb on the bSite
image

Njpatman:
> Sorry I didnt mean pointing as in rotation, but rather that both triggers seem to be trying to 'access' the A site Through some more testing just now I found out that it works fine if the bSite is blown up first and then the aSite second, however if the aSite is blown up first and the bSite second the following error pops up when trying to plant the bomb on the bSite image

yeah, this is a known issue, doesn't break anything so it's gonna stay for a bit till I can get around to fixing it

Pi123263:
Oki, so I assume both sites are fine from a gameplay perspective if they spawn correctly, or?

Njpatman (approved):
Yup, you're good to go. All looks good

PR #50, opened 2024-04-24
Update Silo Security to fit more players, Add Silo Search S&D variant

Pi123263:
Does what the title says
Updates Silo Security to fit more players
Adds S&D variant for better gameplay


Njpatman (approved):
Looks Good

PR #61, opened 2024-05-01
Dagger opdrew

MrPvTDagger:
Rebalance and cleanup


tanaaKa:
Did you want to play this at a CFC? It'll need more slots if so.

MrPvTDagger:
Kinda, but the more I think about it I don't think the mission fits it.

tanaaKa (approved):
lgtm

PR #67, opened 2024-05-04
Add mission file + EdenMissionTemplateBase

Geek-x104:
EdenMissionTemplateBase includes the AIWorld for Eden and the Lobby_CRF_Mode entity pre-placed alongside a series of layers for ease of mission making on Eden. I can make more of these for Arland etc. if needed.


tanaaKa (commented):
I think this would greatly benefit from russians using Mi8s (rather than BTRs) and maybe a captured Zu-23 truck on blufor. What do you think?

Geek-x104:
I agree, I'll get on this now

tanaaKa (approved):

PR #74, opened 2024-05-14
Add adjustable bomb site object

tanaaKa:


JessevGool (commented):
LGTM, can test in a few hours if no one has done so yet

Geek-x104:
Tried changing the object to a replacement (fully set up identically to the original just with a different model) and the bomb is no longer appearing.
The S&D object in the Lobby entity is fine.

tanaaKa:
What object?

Geek-x104:
Originally "GeneratorAirfield_US_01.et", set up a different object as that one doesn't show default in the parent context list.

Regardless, if you try to copy the link to any object from the world editor, it doesn't work, I am now testing by reverting the object to it's original and manually entering the name of the new object.

tanaaKa:
You can't just paste a prefab entity name there. It needs to be the resource path via "Copy Resource Name(s)"
image

It should come out to something like `{3E562E27A2B86F47}Prefabs/Structures/CRF_Bomb.et`

Geek-x104:
Tested it and it worked, the bomb changed however now there is no plant or detonate option, gimme 5-10 to troubleshoot this however overall it looks pretty good to me. (i'm chalking it up to user error on my end if people find/are finding that they are able to set up new entities fine)

tanaaKa:
Yep, changing the entity will involve setting up a new entity with the appropriate bomb plant/defuse action assigned in a location relevant to them.

tanaaKa:
Also someone needs to approve

Geek-x104 (approved):
look good

Pi123263:
Bröther I request ~~lämp~~ the use of stringtable entries
I'll just throw that in in a later language commit

PR #84, opened 2024-05-22
Feature: Custom vehicle: Mountable BTR-70 variant (Base, USSR, FIA, and AFRF versions)

GHPCee:
**Mountable BTR-70**
BTR-70 for usage by mission makers (and in the future, game master) that offers an additional 12 seats on top of the hull.

Inheritance was set up in a manner that allows for this vehicle to fully function with [FPTR](https://reforger.armaplatform.com/workshop/607D472207F3EE65-FPTR) along with any other QOL mods created for the vanilla BTR-70 by other modders.

After testing, live mission usage, and feedback, I'll make this as my first standalone mod for the Reforger Workshop after that new update officially comes out.

- Seats:
  - Crew: 2 | Commander/Passenger: 1 | Passengers: 19
  - Total: 22

btr_mod_5_crop
btr_mod_2_crop
btr_mod_1_crop
btr_mod_3


tanaaKa (commented):
Are you sure he need three different variants? In this case it's only factions which only affect the zeus interface right? 

Vehicles aren't limited by which faction can operate them.

Edit: I guess the camos are different, but that's all I can tell.

GHPCee:
> Edit: I guess the camos are different, but that's all I can tell.

Three variants purely for 3 different camos. Base, FIA, and AFRF are the only three possible variants with the existing textures and mods we run.

tanaaKa (approved):

PR #87, opened 2024-05-23
Fix BW Loadouts

Pi123263:
This PR:
-Adds G36 Variant with XPS3 Holosight
-Adds unsuppressed MP7 Variant
-Gives BW Characters those weapons
-Gives BW AMG Binocs


Njpatman (approved):
Looks Good

PR #88, opened 2024-05-24
Add Escort Service Mission

Pi123263:
This PR will:
Add the Escort Service Mission
Short Description:
Blufor has to get fuel trucks from spawn to obj location
Opfor has to kill blufor and capture at least one fuel truck ( out of 2)
Setting: 1980s


Njpatman (changes requested):
image

Gonna need ya to fix the floating fuel truck in blufor spawn

Also, set the AO border dots to black instead of red, makes them much easier to see.

Njpatman (approved):

PR #99, opened 2024-05-28
Add 'Two Front War' Mission

Pi123263:
Short Description;
OPFOR has been holed up on 2 Islands on Worthy Island connected by bridges, BLUFOR has to attack from both sides (depending on player count) and wipe them out/capture the islands

Notes:
If there arent enough players game master might have to move parts of blufor to the rest of the players so they can attack from one direction as one force

Factions Used:
US/USSR


tanaaKa:
Also lacking objective entities

tanaaKa (approved):

PR #129, opened 2024-06-15
add MMG and Medic teams

Geek-x104:
as per title, adds USMC Medic team (same as 1980s) and an MMG Team (also same as 1980s).


RandAlThor1:
QA Feedback

Medic team works fine.

MMG teams causes error when added to the mission and doesn't appear properly on slotting screen after trying to ignore said errors.

image

image

Geek-x104:
aight i'll remove it and just push the medic team, i was having trouble with the mmgg team regardless

Geek-x104:
@RandAlThor1 done, mmgg team removed, dunno why it had so many issues but it should be sorted now.

RandAlThor1 (approved):

PR #138, opened 2024-06-20
changed Bundeswehr radio code

Arkangel-Vigilamus:


RandAlThor1 (approved):
in future iterations of this mission or ones like it I'm gonna require the OBJ's to be scripted.
but as BW is likely being removed soon, I just want to see this get played.

- will need to verify radio changes on server, can't really test alone

PR #141, opened 2024-06-21
Fixed an issue with detecting factions in the airfield.

Salamiboio:
Fixed an error in the code that made the GetCharacterCountByFactionInsideTrigger function get all factions in the trigger rather than just the USSR faction.


RandAlThor1 (dismissed):
Something tells me you looking at the safestart code and using how it determines OPFOR vs BLUFOR vs INDFOR would be useful for making this code better for a gamemode rather then a specific mission

RandAlThor1:
also, stop committing resoureceDatabase.rdb

Salamiboio:
my b :(

PR #142, opened 2024-06-21
Stop tracking resourceDatabase

JessevGool:
Fixes the resourceDatabase still being tracked. 
If you add a file to the gitignore that is already being tracked it won't just stop tracking it.

Ran the following command to remove the cached file from tracking
`git rm --cached .\resourceDatabase.rdb`


MrPvTDagger (approved):
beat me too it, learnt something new. 😄

PR #151, opened 2024-06-24
Add Experimental Platoon Prefabs

Pi123263:
This PR adds:
-Platoon Prefabs for US80s
-> 1st Platoon (Infantry)
-> 2nd Platoon (Infantry)

-Platoon Prefabs for USSR80s
-> 1st Platoon (Infantry)
-> 2nd Platoon (Infantry)

The thought behind these is to take out some of the repetitive work when making missions of placing down, renaming, etc. platoons/squads.


RandAlThor1:
QA Feedback

I'm all for the creation of platoon prefabs, though I believe "infantry platoon" "support platoon" ect ect are probably more useful and descript, and don't create a false sense of security so MMs still check their lobbies 

This also ties into my next step of trying to standardize factions. I'm going to reach out to mission makers to see what kind of group prefabs they would like to see

**US 1980s**
- First platoon: Squads are in the wrong order 
- second platoon is still labeled 1PLT, 1-1, 1-2 ect
- when you place both down it can get pretty wonky (see image one)
- they don't iterate properly making the point of a second prefab kinda useless (see my opening statement)

**RUSSIA 1980s**
- same issues as the US

image
image

Pi123263:
Reducing it to 1 generic infantry platoon and adding other variants definetly makes sense, just gotta experiment a bit with adding vehicles but should be simple enough I think

The callsigns (1-4, etc.) are afaik all done by the faction manager in a mission and not by these prefabs, these only do the group names (4th quad etc.), if I understand the current system right

RandAlThor1:
Closing this as its dead in the water. create more group-prefabs in future

PR #152, opened 2024-06-24
Fix Faction Manager default callsign values

Pi123263:
This PR changes the default callsigns used by the faction manager prefab created using the crf lobby gamemode prefab.
Should now be standardised to the names used by us.

Additionally there are also some smaller fixes with radio encryption keys/random assignment

### DIFF
--- a/Prefabs/MP/Modes/Lobby/Lobby_CRF_Mode.et
+++ b/Prefabs/MP/Modes/Lobby/Lobby_CRF_Mode.et
@@ -83,10 +83,160 @@ PS_GameModeCoop : "{0F307326459A1395}Prefabs/MP/Modes/GameMode_Base.et" {
   SCR_FactionManager "5EA8C932C68D400F" : "{4A188E44289B9A50}Prefabs/MP/Managers/Factions/FactionManager_Editor.et" {
    ID "60E1584A0994EEB9"
    Factions {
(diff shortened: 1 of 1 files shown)


tanaaKa:
You added group names, not callsigns. Callsigns are 1PLT, 1-1, 1-2, MAT1, etc... I'll make a comprehensive list here in the pr momentarily.

Also every mission is going to be different based on how the mm sets up their structure. I'm not sure if this is beneficial or not because someone still has to go in the fac man and change callsigns to match their platoon or company.

Pi123263:
Oh derp, will fix in a bit

I think its beneficial since 2 platoons with 2 infantry squads seems like a relatively common starting point and also fits in with the other PR #151 that adds prefabs for entire platoons, but I'm biased based on how I make my own missions of course

And if its different for some mission makers they'd have to go into the faction manager to change it either way

tanaaKa:
Right but if there's two elements of something, in this case platoons, then there needs to be a coy element to handle them.

We don't currently have the numbers to make a company element make sense.

RandAlThor1:
The majority of missions don't run 2x infantry platoons and the ones that do almost always have one of the platoon elements locked off. 

while I'm all for the streamlining of groups to make mission making easier I don't think this is currently in scope for our size.

Pi123263:
okih

PR #162, opened 2024-06-28
[Feature] CRF Briefing Screenshots

Geek-x104:
adds briefing images to CRF, if we need to reuse them we come here, saves looking for them elsewhere. New ones added every so often.


RandAlThor1:
I've been thinking about this. would you perhaps be more interested in creating a discord thread with all of the AAR mission posts including their briefing img. that way we can just copy all of the text and the imagine from the thread rather then just having the image in the mod.

tanaaKa:
@Geek-x104 Read above

Geek-x104:
> I've been thinking about this. would you perhaps be more interested in creating a discord thread with all of the AAR mission posts including their briefing img. that way we can just copy all of the text and the imagine from the thread rather then just having the image in the mod.

@RandAlThor1 agreed, let me know where you want it and I'll do it when I wake up in the morning

Geek-x104:
I'll put this in #development as `cfc briefing screenshots`

Geek-x104:
Done, will close this

PR #163, opened 2024-06-28
1.2 update

Arkangel-Vigilamus:
Replaced old slots with shiny new slots
Replaced M113s with 4x Humvee w/50s.


RandAlThor1:
QA Feedback

- squads on both sides need to be renamed.
- also either add two more armor teams, or delete 2 Humvees

RandAlThor1 (approved):

PR #170, opened 2024-06-28
[Mission] Operation Mavrick

MrPvTDagger:
Mission Pull Request Template

This Pull Request will:

Add the mission <mission name!> made by <author name!> to the repository.

- [x] Mission File
- [x] Side-based briefings
- [x] Objective Entities
- [x] Mission .conf file

**Faction(s):**
BLUFOR - Defending
- [ ] US
- [x] RHS_US_USMC
- [ ] CDF

OPFOR - Attacking
- [ ] USSR
- [x] RHS_RF_MSV

INDFOR - Attacking/Defending/NA
- [ ] FIA

Opfor destroy both objectives
Blufor defends at least one objective

**Terrain:**

Serhivvka Ukriane

Opfor targeting drone boosters to allow their tanks to push through the region.

**Slotting Ratio:**

OPF 3:2 BLU

**Time limit:**

60min

### DIFF
--- a/Missions/CRF_SnD79_OpMavrick.conf
+++ b/Missions/CRF_SnD79_OpMavrick.conf
@@ -1,0 +1,13 @@
+SCR_MissionHeader {
+ World "{E5C83D40CC7A70DC}Worlds/Lobby/Serhivvka, Ukraine/Dagger_S&D_Op_Mavrick.ent"
+ m_sName "CRF S&D82 Operation Mavrick"
+ m_sAuthor "Dagger"
+ m_sDescription "OPF 3:2 BLU | RU MSV targets ukriane drone boosters"
+ m_sIcon "{B7385DA4504BD4D5}CRF Images/crflogo.edds"
+ m_sLoadingScreen "{B7385DA4504BD4D5}CRF Images/crflogo.edds"
+ m_sPreviewImage "{B7385DA4504BD4D5}CRF Images/crflogo.edds"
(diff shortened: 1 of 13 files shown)


RandAlThor1:
QA Feedback

- we are trying to avoid directly referencing the Ukraine conflict when it comes to public missions.
- OPFOR slotting is mis ordered, PLT needs to be at top, and LOGI at the bottom.
- BLUFOR logi team needs to be renamed, and PLT needs to be at the top.
- the LOA lines seem off. the blue line indicates how far blufor can go during SS, and the red line OPFOR
- still looks pretty daylight to me

image
image

RandAlThor1:
QA Feedback

Currently this map is broken, nothing you can do about that but just a heads up gonna have to wait until this map gets updated to play it.

- There should be an assets part of the mission description so both sides know what to expect in terms of vehicles in the mission.
- I'm still not sold on the safe start LOA's, there is 100% scenario's where they both see eachother at safestart, and it also forces opfor into a field crossing into dugin blufor when they don't have the assets to do that. (see image for roughly where I think they should be)
- Squads need to be numbered 1plt, 1-1, 1-2, 2plt, 2-1. 2-2 ect. this can be done in the faction manager.
- opfor still has weird ordering it can be a bit finiky.



image

RandAlThor1:
Thank you for putting in the effort for making this mission. 
unfortunately this map is being removed as in its current state its unplayable.

PR #210, opened 2024-07-21
Fia inherit rework

RandAlThor1:
Rebuilds all of the FIA character and group prefabs.
Total of 37 character prefabs and 12 group prefabs

Character prefabs contain 3 Base prefabs.

Rifleman, Leader, and Crew.

All other prefabs are clearly labeled which base they inherit from.


tanaaKa:
Isnt having three different base prefabs counterproductive to having a base prefab at all?

RandAlThor1:
> Isnt having three different base prefabs counterproductive to having a base prefab at all?

My concern was if we wanted to change UGL ammo count we would have to do it to all of the leader slots. And having the Leader radio would prevent changes to the uniform happening properly.

And with crew they operate entirely different weapons and uniforms. 

I can change it to a single prefab

RandAlThor1:
Also, specific to FIA their AK-74 for leads with the UGL isn't compatible ammo wise with the normal rifleman weapon. so having them a single base prefab would mean all leader slots would need to be updated separately anyways

tanaaKa:
Hmmmm. I'm kind of torn tbh. Wouldnt a FTL prefab become the leader "base" then?

Rifleman : FTL : SL/PL (they're the same lol)
Rifleman : AAR : AR

etc...

Crewman base makes sense

RandAlThor1:
I figured a leader base would make sense in case we ever wanted to specifically change FTLs. also color team assignment via CSI we don't want to inherit

tanaaKa:
It's semantics at this point, but wouldn't that double our work when we need to change something then? Having the FTL serve as a leader base prevents work.

RandAlThor1:
maybe? csi was the biggest thing stopping me from using them. SL might make more sense to avoid that

tanaaKa:
Yeah that's biting us tbh. I'm honestly okay with just forcing FTLs to set their own color from the jump. Lets get crew to inherit from rifleman then just overwrite as necessary. 

@Njpatman Is there **ANY** other way to assign color teams via CSI that doesn't require us to have double prefabs for each role and color team? It's super inefficient.

Maybe a server component which we can pass prefab arrays to? I'm not sure.

RandAlThor1:
FTL change will be easy enough.

just to clarify for crew so I don't misunderstand.
do you want me to have all the crew inherit directly from rifleman, or have one crew inherit from rifleman and all other inherit from them

tanaaKa:
Rifleman: Gunner/Driver: Commander

Gunner and driver are the same and should have GI radios, carbine rifles, and crew uniforms.

RandAlThor1:
@tanaaKa TL/driver now both inherit from rifleman, then leaders/crew respectively inherit from them.

tanaaKa:
Only small changes is we should prob rename the prefabs to remove *_Character_* from them imo. I'll get working on that.

tanaaKa (approved):

PR #218, opened 2024-07-26
[Mission] Add First CAH mission

tanaaKa:


RandAlThor1:
QA Feedback


I've Fixed these
image
image
my group prefab was named wrong fixed that as well
image



briefing screenshot for future reference 
image

Issue #220, opened 2024-07-29
[CRF General Bug] UGL Smokes disappear too soon


RandAlThor1:
@Salamiboio 

The UGL smokes pop too soon/don't last long enough when being fired at a longer distance they are gone by the time they land.

Salamiboio:
I'll increase the time they ploom but that's weird, didn't encounter that in testing. Must be new update

tanaaKa:
@Salamiboio Reminder

Salamiboio:
Shit forgot about this will be back to my computer tomorrow will check then

RandAlThor1:
@Salamiboio I am once again asking if you will look into this

Salamiboio:
I can't replicate what you're talking about, in my workbench they work just fine, they deploy at a distance and ploom for awhile.

RandAlThor1:
idk man, go fire them in the server and tell me their fine

Salamiboio:
I don't know then, it's just the flare with smoke particles instead of flares. I think it may just be a server replication issue which is why the BI devs haven't done this yet.

Salamiboio:
I'll look into redoing this not using the flare system and instead just using base game smoke grenades

PR #274, opened 2024-09-20
Make smokes smokier

tanaaKa:
#220


Njpatman (approved):
may the frames be ever in your favor

PR #303, opened 2024-10-17
Harry's first mission

HarryEngineering:
Made a simple mission on Eden called Power Outage.

Mission Pull Request Template

This Pull Request will:

Add the mission eden_power_outage made by Harry to the repository.

- [x] Mission File
- [x] Side-based briefings
- [x] Objective Entities
- [x] Mission .conf file

**Faction(s):**
BLUFOR - NA
- [x] US
- [ ] UK
- [ ] RHS_USAF 
- [ ] CDF

OPFOR - Attacking/Defending/NA
- [ ] USSR
- [ ] RHS_AFRF

INDFOR - NA
- [x] FIA
- [ ] RHS_ION
- [ ] TALIBAN

Blufor holds the power plant for 15 minutes, and then pushes to break through the harbor at St. Pierre with a minimum of 4 players.

**Terrain:**

Eden

Bluefor forces have been tasked to capture, and repair a damaged power plant from local insurgent fighters. Once the power plant has been retaken and repaired, Blufor must push to St. Pierre and break through to the harbor there.

**Slotting Ratio:**

<4:3>

**Time limit:**

<60min>

image
blufor 
image
indfor
image

**Design concept/thoughts:**

Creating a relatively simple mission for my first go. The balance is pretty close, but blufor has more a bit more, and more firepower, as indfor decides how conflict is going to happen. Blufor has a minimum player threshold to prevent too much cat and mouse when the game is running long. 

**Other Notes:**
INDFOR is NOT allowed to detonate C4 in the power plant. Doing so constitutes a loss. 
There are triggers to display messages for capturing the power plant, capturing the harbor, and blufor running out of people.


RandAlThor1:
**QA Feedback**

- Spawn area markers should not be visible to everyone, the indicated boxes need to be toggled off. you can do this on mass by selecting all of the markers and changing it
image

- FIA should have a spawn flag set, you also need to place down a INDFOR safeStartBorder same as you have for BLUFOR to indicate the area INDFOR can move before the mission official begins. example attached. under no one within either sides SafeStartBorders should be able to see the other side
image

- Both sides should have an Ambulance vehicle being the "M997_maxi_ambulance" and "UAZ452_ambluance" you should generally have one per medical officer, as this allows medics nearby to greater treat critically wounded players.

- Objective markers are pretty hard to see, I would prefer they get changed to something more striking for the map, they also need their text changed to which objective they are and not just OBJ
image

- you don't have any Objective entities on the map, these are used to show in the mission end screen what objectives were accomplished and to recap the overall score, you need one per side, per objective.
image

- INDFOR Brief is called OPFOR brief
image

- Your Winning conditions brief should be titled "Win Conditions" and should be reformatted, since there is two objectives, since this should be a brief for both factions, it should state the win conditions for both sides.
image

- if INDFOR is expected to defend this large AO, they will need more transport to get themselves around it effectively. else they will likely abandoned one objective in favour of strong arming another

- the Humvee's from the "Humvee Varients" mod cannot be placed as they will cause errors loading the mission on the server, this is generally a reforger issue. in order to counteract this we regularly use scripts to delay the spawn in of problem vehicles. take a look at "lamentinLimbo" for a very good use of these scripts, you do not need to be nearly as through as that. but effectively any non-vanilla asset should be spawned in using this to ensure the mission loads.

- you need to create a mission config file, these are located in the missions folder, inherit off of "CRF_BaseMissionConfig.conf" and replace the info with your missions info
image

RandAlThor1:
**QA Feedback**

- the group names need to be something like "1st squad" or "1-1" ect depending on the scheme your going for, since the left most name is what appears on the minimap
image
image

- no reason to lock INDFOR in so much during safe-start as the defenders they need free reigns to positions themselves around the objectives. I've marked out a line that would probably be better
image

- your trigger for indfor player count triggers even with 4 INDFOR present, I know from us doing this in the past. game-state and replication can be an issue. Salami has similar code in his Supply Extraction gamemode, I recommend looking into that or talking to him about it https://github.com/CoalitionArma/Coalition-Reforger-Framework/blob/release/scripts/Game/GameMode/Supply%20Extraction/CRF_SupplyExtractionGame.c

- power and harbour triggers worked well in the peer tool for me, your Harbor trigger is updating every second. bring that inline to be every 15 seconds+ like the PowerTrigger, it can hurt on the server a decent amount to update that often.

- your two BLUFOR objectives aren't appearing in the debrief screen. they just need their faction key capitalized
image

HarryEngineering:
>     * your trigger for indfor player count triggers even with 4 INDFOR present, I know from us doing this in the past. game-state and replication can be an issue. Salami has similar code in his Supply Extraction gamemode, I recommend looking into that or talking to him about it https://github.com/CoalitionArma/Coalition-Reforger-Framework/blob/release/scripts/Game/GameMode/Supply%20Extraction/CRF_SupplyExtractionGame.c

I will change over to this counting method, but the intention is the trigger only counts BLUFOR as it's almost certain they couldn't do their objectives if they get so badly walloped. However, if BLUFOR is winning then there is a time limit on the end of the game, and also there is still a possibility that the scraps of INDFOR could accomplish their goal, which would be fun to watch.

RandAlThor1:
I misread what faction was doing what in the trigger for count. retested in peer tool and it should be working. QAing during my day job can be a bit of a mess eh

RandAlThor1:
**QA Feedback**

- just noticed that the time limit is set too 60 minutes. given the size of the AO up it to 75/80 minutes and I will accept it. we generally don't want the time limit to be the reason the mission ends.

good mission overall, like the idea of needing to hold an OBJ for a set amount of time. should be fun

RandAlThor1 (approved):

PR #312, opened 2024-10-24
update mission Schooners end sortie and add USSR 80's KLMK variant

FluffSquirrel:
the exact same mission as it was before, but with klmk soviets instead of khaki, using the gearscript now

KLMK soviet faction isnt complete, I don't know if I'm expected to flesh out every role for it or just the ones I need for the mission I'm making it for. I can finish it if needed, shouldnt take long.


RandAlThor1:
**QA Feedback**

- you have higher than normal ammo counts for rifles, In the config the magazine count doesn't count the one already in the gun, so you would have 10+1, and 5+1 for the UGL HEs instead of the normal 9+1 and 4+1. 
image

PR #318, opened 2024-10-29
Mission updates

HarryEngineering:
Updated Oops, all grenades, and Power Outage.


RandAlThor1:
I'm going to assume the mortar being out there is so that it could be in range of the OBJs.

Salami has created a crew served weapon system mod that allows us to have people move the mortars. you can now just give a gunner and a loader the tube and tripod prefabs. and they can place and assemble/disassemble the weapon at will, I would suggest using this system instead of the awkward mortar in the middle of the AO.

HarryEngineering:
Added crew served mortars, and added a more stringent check for blufor count.

RandAlThor1 (approved):

PR #323, opened 2024-11-03
Add gearscripts for CDF & ChDKZ

FluffSquirrel:
adds gs for both factions. both factions fully complete.


RandAlThor1:
Already added via a mission

PR #337, opened 2024-11-23
Spectator nv gs

Salamiboio:
- Adds a night mode for spectator, so they, ya know, can see at night.


RandAlThor1:
@Njpatman Need a QA pass on this

Njpatman (approved):
It works in my dedicated server tool, calling it good.

PR #341, opened 2024-12-02
Added "Downhill Fast" Domination Mission

kevinwork98:
This Pull Request will:

Add the Domination mission "Downhill Fast" made by FanServ to the repository.

- [x] Mission File
- [x] Side-based briefings
- [x] Objective Entities
- [x] Mission .conf file

**Faction(s):**
BLUFOR - Defending
- [x] RHS_USAF 

OPFOR - NA

INDFOR - Attacking
- [x] TALIBAN

Capture/hold both objectives, the town of Khan Abad and the MSR to its south.

**Terrain:**

Kunar Province. Northeastern corner, town of Khan Abad, mountain south of Sarab.

A United States Army vanguard force has captured the town of Khan Abad in advance of a convoy from FOB Michigan which will pass through the nearby MSR eastwards. However, Aerial Ground Surveillance has spotted a sizeable Taliban element descending from the mountain town of Sarab towards Khan Abad. If the town and MSR are captured, not only will the MSR be covered with IEDs, the convoy will be sitting ducks at FOB Michigan.

**Slotting Ratio:**

3:2 INDFOR:BLUFOR

**Time limit:**

60min

DownhillFast

**Design concept/thoughts:**

The defending force has cover within and behind several buildings, but they are inside of a deep bowl which the attacking force can shoot into from several angles. There is some terrain which the defenders can maneuver into to gain some elevation, but cover there is limited so they are vulnerable to suppression. To compensate for the disadvantages, the defenders have access to 4 armed Humvees with shields. However, the attackers still have anti-tank kits which can easily destroy the vehicles if they are exposed and static.

**Other Notes:**

Assets:
BLUFOR: 4x Armed Humvees with Gun Shields.
INDFOR: 1x MMG Team

There is a hill to the east of the objective right next to the map border which completely overlooks the town. It seems far too advantageous to the attackers, so I have excluded it from the AO. Must keep an eye out for players trying to take that position. I have also limited the eastern border of the AO so that INDFOR has to stay on the visible western side of the hill in the forest. This is to prevent INDFOR from just sneaking up to the town on the eastern side of the hill.


kevinwork98:
I'm thinking about cutting BLUFOR down to 1 platoon with 3 squads + 1 MMG. That would put the ratio at 34:58. Also considering giving INDFOR an MMG squad and taking Platoon 2 down to 2 squads (not that 2nd platoon will get filled in all likelihood). Altogether that would be 34:50 which is pretty close to 2:3.

But I don't know if MMGs are a good idea for this mission.

Edit: I think I'll just give BLUFOR 4 squads and give INDFOR an MMG on Platoon 1 to help counter the Humvees.

Edit Edit: Also thinking about giving INDFOR some markers so they're able to prepare their approach better during Safe Start. As it is, they only have the vague map to plan an offensive around. Or I could play PL/SL for INDFOR.

RandAlThor1:
**QA Feedback**
- callsigns, (the text in the brackets) needs to be set via the faction manager
image
image


- special teams go on the bottom of the list, besides medical teams.
image

- indfor doesn't need a 2nd platoon, not enough to justify it, plus we never fill it.

- I don't mind the objectives, the AO is too narrow, people will break out accidently and there is no real, natural lines to follow. add more space to the west to give attackers a bigger area to attack from. night vision does make a VERY big difference for these kind of missions.

- indfor is spawning a bit weird, this can cause unintended deaths sometimes. just gotta move them over a bit 
image

- 100 minutes is too long for a mission like this
image

- config needs to point at your world file
image


- overall for a first mission, very solid

RandAlThor1 (approved):

PR #342, opened 2024-12-04
Added "Swing Past" S&D mission to Zimnititra

kevinwork98:
This Pull Request will:

Add the mission "Swing Past" made by FanServ to the repository.

- [x] Mission File
- [x] Side-based briefings
- [x] Objective Entities
- [x] Mission .conf file

**Faction(s):**
BLUFOR - NA

OPFOR - Attacking
- [x] USSR

INDFOR - Defending
- [x] FIA

OPFOR:

WIN: Destroy at least 1 cache and reach the extraction point.
TIE: Destroy both caches and fail to reach the extraction point.
TIE: Reach the extraction point without destroying any caches.

INDFOR:

WIN: At least 1 cache survives and OPFOR does not reach the extraction point.
TIE: Both caches survive and OPFOR reaches the extraction point.
TIE: Both caches destroyed and OPFOR does not reach the extraction point.

**Terrain:**

Zimnitrita

Command has tasked a Soviet Motor Rifle Platoon with moving to the bridge north of the factory at Glochy. Because of their approach vector, command has optionally tasked the platoon with destroying the local weapons caches used by the FIA in the town of Nova Frusca. However, the priority is to preserve forces and extract at the rally point by the bridge.

While there is only a single FIA platoon garrisoned at Nova Frusca, it is known that they are armed with an HMG and Medium Anti-Tank weapons. They will contest an attempted extraction, but their priority is to defend the caches.

**Slotting Ratio:**

4:3 OPFOR:INDFOR

**Time limit:**

90min

SwingPast

**Design concept/thoughts:**

Utilizing the interesting geography of Nova Frusca and the combined Search & Destroy + Extraction objectives, both teams will have lots of interesting decisions to make regarding how to win the game. There are a lot of situations that can end in a valid tie, which both teams will have to go all-out to prevent. This prevents INDFOR from simply camping the extraction and OPFOR from rushing to the extraction without doing the objectives.

**Other Notes:**

OPFOR:
1x Motor Rifle Platoon

INDFOR:
1x Infantry Platoon
1x HMG Team
1x MAT Team
2x URAL Transport
1x UAZ


RandAlThor1:
**QA Feedback**

- alot of the same issues as the last mission which makes sense, callsigns need to be fixed
- AO isn't really gonna be respected to much, too narrow and no reason to respect it. things line mines can be useful for things like this
- personally not a huge fan of the objective as a whole, extraction makes sense as a second objective, but a whole mission about running through a town doesn't seem that great of a concept.
- the extraction point is very close to the main objective, so they will likely still be in contact at it.
- 4:3 ratio doesn't really seem needed here, 3:2 ratio will probably lead to better results
- when using BTRs replace gunners with commanders so that they have long range radios to talk to crew and squad

RandAlThor1 (approved):

PR #362, opened 2025-01-01
Efficency and prefix changes

Njpatman:
- Changed all local coalition lobby CRF prefixes to CLB for easier traceability.

- Combined all client components into one main client component to (hopefully) help network performance. Updated all game modes to work with this system.

- Safestart is now an extension of the base game mode, so any mission made with the CRF_Lobby game mode will have it. This has also proven to help (locally) by deleting the bullets faster so players can't kill each other at close range anymore. Mission makers will need to reset their safe start parameters.

- Game Timer and Safestart replication doesn't happen as often (100ms to 1000ms), and the game timer is only checked every 5 seconds instead of every frame.

- Removed many old overrides that no longer work. This should include all the reforger lobby overrides and WZ turret overrides, as well as any unsupported guns or clothing.


tanaaKa (changes requested):
image

Getting this on launch. Also was renaming it all from CRF really necessary? Seems so nitpicky.

Njpatman:
> image
> 
> Getting this on launch. Also was renaming it all from CRF really necessary? Seems so nitpicky.

CRF_Modded has been deleted since i integrated it into the lobby code, re-pull please:
image

tanaaKa:
The errors continue 
image

Njpatman:
I just like easier traceability in case any scripts crash/throw a bunch of errors, especially since these scripts are so mixed together

> The errors continue image

Did you pull the lobby PR of the same name as well? The two PRs are intertwined because the two codebases are intertwined

tanaaKa:
No you cant pull two PRs at the same time. That's the entire point in a PR lol

tanaaKa:
Have done it manually. Lets just get this merged so we can move on. I think @Salamiboio will need to update the mod on the workshop and add all of us to it.

Njpatman:
> No you cant pull two PRs at the same time. That's the entire point in a PR lol

How do I do a PR across 2 repos?

Issue #370, opened 2025-01-02
Opening inventory crashes you (gearscript?)


tanaaKa:

Salamiboio:
Invis bags caused this

RandAlThor1:
#384

PR #384, opened 2025-01-02
Invisible backpack crash fix

RandAlThor1:
Invisible backpacks had a call to a removed vanilla item, this has been corrected
issue was closed before I implemented the fix, @Salamiboio don't do this plz and thank you


RandAlThor1:
#370 fixes this issue, was closed before PR was made or pushed

PR #406, opened 2025-01-05
Harry SPCL80 TerribleTuesday - Anizay

HarryEngineering:
Mission Pull Request Template

This Pull Request will:

Add the mission Terrible Tuesday made by Harry to the repository.

- [x] Mission File
- [x] Side-based briefings
- [x] Objective Entities
- [x] Mission .conf file

**Faction(s):**
BLUFOR - Defending
- [x] RHS_USAF 

INDFOR - Attacking
- [x] TALIBAN

Blufor wants to distribute the supplies to the locals, and make it to exfil. Taliban would like to stop that.

**Terrain:**

Anizay

A silly little RP mission.

**Slotting Ratio:**

<2:1>

**Time limit:**

15 minutes

image

**Design concept/thoughts:**

Based on an older Arma 3 mission, just brought to Reforger.

**Other Notes:**

Safestart has been left on for people to orient themselves, but I think instantly readying both sides is good. 30 seconds should be plenty.


RandAlThor1:
**QA Feedback**

- not all of your walls are snapped to the ground
image

- reading through the briefings, I am confused as to how exactly supplies get given out and what the scripted exit is for?

- we have had similar RP esc things like this in the past in A3, with 30ish people we found that a slightly larger down with some trucks and stuff was good for the numbers still, so probably scale things up a bit

HarryEngineering:
Ok, I moved to Khas, removed the confusing exit trigger, clarified the supply language, and scaled the special up a little.

RandAlThor1:
**QA Feedback**

- I much prefer this AO, im still confused as to how BLUFOR is supposed to accomplish their objective, whats stopping them from just dumping the supplies and leaving? how could indfor sabotage their efforts?

- otherwise looks good

HarryEngineering:
My initial thought is that things would break down pretty quickly because it's a special, and not need deeper systems, but I could see a world where nothing ends up happening. I changed blufors objectives to have a reason to stay in the AO, and made the supplies more of a bartering chip with civilians who have their own objective now. I think this is much more dynamic.

RandAlThor1 (approved):

PR #412, opened 2025-01-10
Prejoin 0.5

Geek-x104:
Adds a rudimentary prejoin until lobby has spawn points


RandAlThor1 (approved):
not much of a prejoin really, we need to build out systems similar to A3. aka teleport poles ect to allow for things to do.

respawn system would also be very helpful

PR #414, opened 2025-01-10
Feature/fuckingstats

tanaaKa:
Adds basic implementation of stat tracking V1. It only adds an entry to the players collection on player connect.

This should give us a basic understanding of performance with mongo db.


Njpatman (approved):
Looks good, if hintmanger causes issues, it does.

RandAlThor1 (approved):

PR #415, opened 2025-01-10
Add faction order

### LINKED ISSUE #407: [CRF General Bug] Placing units in game master spawns a random unit
Usually, it happens with vehicles, caused by factions having the same order index.

Can be fixed if we set each faction in our faction manager to a different order index.

tanaaKa:
Fixes https://github.com/CoalitionArma/Coalition-Reforger-Framework/issues/407


Njpatman (approved):
Looks Good

PR #418, opened 2025-01-11
AA Team Fixes

### LINKED ISSUE #417: AA Teams Only Have 1 Rocket, Need to Increase To 3-4 For Balance
With the addition of flares to WCS vehicles, we should increase the amount of ammo the AA teams have to make them effective.

Njpatman:
Fixes: #417 

Just config changes

Please check the GUIDs in each GS to see what was updated; some mods renamed their magazines.


RandAlThor1 (commented):
Imma just fix this myself, but giving everyone an even bigger alice pack aint it.

for now, assistants just need launchers so well just make the launchers smaller and give them a few. that being said, I really don't want assistant AA being used while they are not able to reload. I'd rather just multiple rifleman AA

Njpatman:
Aight, fair enough, good luck o7

PR #425, opened 2025-01-12
Add Respawns

MrPvTDagger:
Adds respawn system


Njpatman (changes requested):
Why are we changing the filename of the admin menu?

https://github.com/CoalitionArma/Coalition-Reforger-Framework/pull/425/files#diff-abc9929955523f76549c2a9886d8427a3b453c997e4f91abeeb8c2819a29031bL9

Njpatman (approved):
Ran Fine in Dedicated, no errors, and @Salamiboio gave his blessing. so we ball

PR #439, opened 2025-01-15
Adds Respawn Pole and Teleport Pole System

Njpatman:
Instead of manually naming spawn points, mission makers can now drag and drop spawn points from the resource browser. This also opens up the ability to enable/disable respawn points based on scripts in the mission and mobile respawn points can now be made.

Respawn point Priority hasn't been implemented yet. The option exists for now but doesn't contribute to where a player respawns.


Njpatman:
Whoops, all good to go now.

PR #469, opened 2025-01-17
Fixes Bradley TOW

Salamiboio:
Fixes Bradley TOW by replacing it with a kornets rocket


Salamiboio:
Are we able to get this on the server before session?

Njpatman (approved):

PR #500, opened 2025-01-27
Squad Leads see their own markers

### LINKED ISSUE #475: [CRF General Bug] Leaders of groups cannot see their group icons on the map
It makes it infinitely more challenging for a group leader to coordinate his group since he has to keep asking other people where he is on the map.

MrPvTDagger:
Fixes #475


Njpatman:
This will only show the squad leader's markers to one side, correct? (you have tested this, right?)

MrPvTDagger:
Didn't test but the faction affiliation is checked in a different function

Njpatman:
Alright, please test in a dedicated server environment with 2 peer tools, each on a different faction, and report back.

Or I'll do it when I have the time today (having another admin/reviewer test would also be acceptable)

MrPvTDagger:
Confirmed on peer tool.
example

Njpatman (approved):

Issue #490, opened 2025-01-19
[CRF Feature] Put player in correct radio freq when they respawn


MrPvTDagger:
I'm sure we can get the radio freq from the group manager and set it on respawn

MrPvTDagger:
added in #515

PR #515, opened 2025-02-02
Feature/respawnwithcorrectradiofrew

MrPvTDagger:
Adds #490 
In testing it didn't mess with spectator voip but I'm worried it will


Salamiboio:
it wont.

Njpatman (changes requested):
Please remove the changes to the rush mission, This PR is only for setting radios on respawn

Njpatman (approved):

PR #557, opened 2025-03-16
Add gearscripts: CDF (Harvest Red), ChDKZ (Harvest Red), (CDF) Modern

FluffSquirrel:
adds:
harvest red CDF and ChDKZ gearscripts
modern CDF with m4s
nerfed 6b3 rig that isnt OP like the vanilla one
2 RHS helmet spartan-3 variants with NVGs for use by modern CDF
removes depreciated cdf, chdkz gearscripts

issues:
- 6b3 rig doesnt passthrough damage when wearer is uncon, making doubletaps harder not sure how major this issue is? 1-2 shots to a limb, a shot to the head, a bayonet stab will still kill them. not exactly sure how to fix it, i can go back to the drawing board on nerfing the 6b3 rig but theres little documentation on it. Or just switch to 6b2 
- chdkz pants, m88 hat need to get updated in base mod, will automatically fix when updated
- chdkz flag needs to be added to base mod, then updated in chdkz gearscript
- Modern CDF are just a reskinned USMC, and im not very sure how to differentiate them to make them feel unique. if anyone has ideas lmk


FluffSquirrel:
Fixed issues 2 & 3

RandAlThor1:
**QA Feedback**

**Modern CDF**
- make sure to duplicate not inherit off of existing gearscripts, if we were to change anything in the USMC gearscript that you didn't want changed for the CDF one we wouldn't know until we encountered it in-game
- in the USMC gearscript there are some rolespecific overrides, alot of these are just for backpacks. some for crews/pilots are for shirts to make them look distinct. because of this the CDF vehicle crews have the same camo as the USMC
image
image
- im not 100% on the modern CDF having M4s, tho not too too many options right now. once 1.3 adds the modernized AKM might be a good idea to give them that.
- otherwise this is just the USMC gearscript which is fine. 

**RED HARVEST CDF**
- same issue with inheritence with the ussr 80s gearscript. love that your using it as a base, but duplicate not inherit to avoid unintended changes.
- love the random helmets. I would just leave everyone getting only proper helmets and override for upper leadership with the caps, armor is important for people in gunfights
image

**ChDKZ**
- same inheritance issue. otherwise zero problems

FluffSquirrel:
Deleted CDF, ChDKZ, Modern CDF
Readded CDF and ChDKZ Harvest Red as duplicated gearscripts. Modern CDF not readded, will be done when 1.3 comes and theres enough variety to justify. 

Issues: ChDKZ icon washed out, will be autofixed on switching import method on base faction mod
Dont know of anything else

RandAlThor1 (approved):

PR #562, opened 2025-03-18
New Mission- Harry SE Lignum Vitae

HarryEngineering:
Mission Pull Request Template

This Pull Request will:

Add the mission Lignum Vitae made by Harry to the repository.

- [x] Mission File
- [x] Side-based briefings
- [x] Objective Entities
- [x] Mission .conf file

**Faction(s):**
BLUFOR - Attacking
- [x] US

OPFOR - Attacking
- [x] USSR

Compete to gather 2000 supplies, and then defend your cache for 10 minutes.

**Terrain:**

Gulfcoast Islands

Opfor and Blufor are competing for supply in this AO. Each side has a small shipping container to fill with 2000 supply. There are main caches labeled on the map with 500 supply in them, and unmarked smaller caches dispersed around the AO features. After a side has filled their shipping container, they will have to defend it for 10 minutes to win.

**Slotting Ratio:**

<1:1>

**Time limit:**

<45min>

LignumVitae

**Design concept/thoughts:**

A couple of ideas here. Using RTO and as such SL's have GI radio. COY and PLT still have dual channel radios. Just trying out the idea, to try to get people talking better. Secondly, I've incorporated recon with plain rifles to try to encourage more information gathering. Thirdly, this is reviving the supply extraction game mode, but in a competitive format. In the unlikely case that one side gathers enough supply without a real confrontation, they have to defend their cache for 10 minutes with a global mark letting the other side know where it is.

**Other Notes:**

There's a script watching the amount of supply in the crates, and will send a message on being filled.


RandAlThor1:
**QA Feedback**
- OPFOR recon is missing an RTO and has two spotters.
- since recon and MMG are not duplicated they would be under COY and not PLT so they should be at the bottom
- realistically I would aim for force parody here, the structure gets a bit weird replacing the OPFOR 2-3 with a BLUFOR MMG team and a Vehicle crew
- you have a few things in your gearscript that are blank, not a huge deal just throw some errors
image
- I get the intent of giving COY/PLT a leader radio as an override. but this results in them being given 2 radios. I would say its honestly worth trying to try using just RTOs 
image
- I would probably also put a marker on the supply depot for both sides. the BLUFOR one is fairly obvious but it took me a moment to find the OPFOR one in particular.
- both sides are missing ambulances
- might be wise to also give vehicles for infantry to transport since the trucks don't have space for players as they are full of supplies even when empty

HarryEngineering:
>     * I get the intent of giving COY/PLT a leader radio as an override. but this results in them being given 2 radios. I would say its honestly worth trying to try using just RTOs

Do you feel strongly about adding RTO to command? I don't think people will slot it, and then that makes COY/PLT's life much harder than 2 radios.

HarryEngineering:
- Changed spotter to RTO
- Moved recon
- The idea for mmg was it asymmetrically balanced the general infantry imbalance. I think it's fine to not use because Opfor has the rpk instead of pkm. Similarly, the vics were asymmetrically balanced, but it's a small issue with only one humvee. Fine either way. 
- Removed the blanks I found in this, and the gearscripts I duplicated from
- Command only has the one dual channel radio now per our discussion
- Put cache markers down
- Put down ambulances
- Fixed trucks having boxes in them

RandAlThor1:
**QA Feedback**

- not a you issue, but right now RTO's are getting GI radios from when they are enabled as well giving them two radios
- I am confused why you want to use GI radios and RTO at the same time, i would suggest just leaving it to upper leadership with radios, and RTOs as the only squad level radio otherwise its too easy to lazy people to bypass the RTO system
-  everything else seems good to go

HarryEngineering:
My goal is to seperate short vs long range communication. It's an experiment for sure. It's easy enough for the SL and RTO to exchange radios if they want, but I hope SLs use the opportunity to share the burden of leadership. SLs communicate directly with their FTLs, and RTOs communicate with other elements RTOs or command. I think what I might do is remove GI radio, and give them back to FTLs via gearscript, the same I did for leadership. I think this will improve the SL experience, create more step ladders into leadership, and more directly engage the FTLs which have been overlooked imo. It'll also fix the RTO double radio.

RandAlThor1:
works for me, once that's done I'll approve it

HarryEngineering:
Removed GI radios, readded to FTLs and Spotters. Also changed command dual channel radio to 2ic and gave vic leads a 2 channel.

RandAlThor1 (approved):

PR #568, opened 2025-03-20
Add a bomb site marker toggle to the search and destroy game mode

HarryEngineering:
Added a toggle in the search and destroy game mode component that's set to false by default, which hides the map markers when toggled. Tested on Gravity, but please make sure I didn't mess anything up.


tanaaKa (approved):
If I'm being extremely pedantic, you should always use positive logic. IE var named `showMapMarkers` becomes

```
if (showMapMarkers) {
  do thing
}
```

Looks fine though.

PR #584, opened 2025-03-31
Make AI Invincible During Safestart

Njpatman:
Pretty self-explanatory. Make AI invincible during safestart so roles cannot be accidentally deleted by a ND


tanaaKa:
Do we need to account for the entity being invincible when someone JIPs into it?

Njpatman:
> Do we need to account for the entity being invincible when someone JIPs into it?

Nah, since invincibility is based on the base character entity JIPs don't matter since they are just controlling the base character entity.

And invincibility is handled only by the server so there's no way for locality to get involved.

tanaaKa (approved):

PR #607, opened 2025-04-28
Backend Overhaul

### LINKED ISSUE #531: [CRF Feature] Make slottable entities invisible until a player takes that slot
Would need to also disable colliders and nametags

Njpatman:
Fixes: #531


Njpatman:
@tanaaKa Need to fix the JIpError I'm getting, but feel free to go through everything and comment/refactor as you see fit.

PR #614, opened 2025-05-12
Add support for multiple spawn points per faction and allow players to select where to spawn

MrPvTDagger:
screenshot


tanaaKa:
Is this ready for review?

MrPvTDagger:
Not yet
- Need to figure how to update the UI when the point is disabled/enabled on the fly and handling keeping what spawn was selected before the change.
- Fix map panning and see if i can draw a marker when it selected (Couldn't get this to work without re opening the map)

Njpatman:
> Not yet
> 
> - Need to figure how to update the UI when the point is disabled/enabled on the fly and handling keeping what spawn was selected before the change.
> 
> - Fix map panning and see if i can draw a marker when it selected (Couldn't get this to work without re opening the map)


For the first one I'd throw a rpl component on the respawn flags and just do a replication.bumpme() to update the flag status for all clients (can also do rplsave and rplload that's updated with a broadcast, pick your poison really). Then have a script invoker/a OnRpl function in the rpl prop attribute to update clients even when they have the respawn menu open. I'd just save the last selected entity on the player controller component

For the second I'd try using the scripted markers that I built, can also manually call the marker if needed.

No clue about map panning tho, may have to hijack the map on the client and make it move via the map input controls and their linked functions.

MrPvTDagger:
> > Not yet
> > 
> > * Need to figure how to update the UI when the point is disabled/enabled on the fly and handling keeping what spawn was selected before the change.
> > * Fix map panning and see if i can draw a marker when it selected (Couldn't get this to work without re opening the map)
> 
> For the first one I'd throw a rpl component on the respawn flags and just do a replication.bumpme() to update the flag status for all clients (can also do rplsave and rplload that's updated with a broadcast, pick your poison really). Then have a script invoker/a OnRpl function in the rpl prop attribute to update clients even when they have the respawn menu open. I'd just save the last selected entity on the player controller component
> 
> For the second I'd try using the scripted markers that I built, can also manually call the marker if needed.
> 
> No clue about map panning tho, may have to hijack the map on the client and make it move via the map input controls and their linked functions.

I tried using the scripted markers but they don't "spawn" on the map if its already open, you have to close it and reopen it
- What I will try is spawning the markers before opening the map instead of adding them when the spawn is selected by the player.

For the pan the code is there i just couldn't figure out the vector:
https://github.com/CoalitionArma/Coalition-Reforger-Framework/pull/614/files#diff-2764ba4584864cf738b91fea7f636960eb1fb779ab9134f21156789fc8d18ad7R342

I think getOrigin is wrong cause in the example i found it used getWorldPos() or something like that but it wasn't available

MrPvTDagger:
Just needs a cleanup and minor issue were initial markers doesn't get cleaned up

MrPvTDagger:
@tanaaKa @Njpatman Ready for review

PR #615, opened 2025-05-13
Add new Ticket System and Redesign of admin menu

MrPvTDagger:
- Added new ticket system flow for admin messages
- Redesign admin menu
- Moved individual menus to individual widget layouts
- ~~Added dedicated admin chat (/ac )~~


tanaaKa:
Why `/ac` instead of `/a`?

MrPvTDagger:
> Why `/ac` instead of `/a`?

Its just for admins/mods to talk to each other so were arent spam making tickets talking to each other

tanaaKa:
Shouldnt we just add a check to `/a` to check if they're an admin? That way there's no confusion. Obv an admin wouldnt create a ticket.

Also get your conflicts resolved and I'll take a look at the code.

MrPvTDagger:
Fair will get that changed.

tanaaKa:
Feedback:

- Default state for the admin menu is a blank "home" screen
image
Clicking on the home button shows these widget(s)

- Hint to specific factions is broken. It always shows to everyone.

Otherwise great work. Much cleaner system.

tanaaKa (approved):

PR #616, opened 2025-05-14
Corrected SOV_AMMO_Smoke_Purple's smoke effect to Smoke_Effect_Purple

wakelagger:
First PR here, so let me know if there is workflow for non-mission PRs I should be following.

Investigating the UGL smoke grenade behaviour, and I noticed that the soviet purple vog's submunition was misconfigured, so I corrected it.


tanaaKa:
Yeah our submunitions have been misconfigured since conception I think. Havent had time to fix yet.

Njpatman (approved):

PR #623, opened 2025-05-22
Wait For Full Gear Initilization Before We Initilize Player

Njpatman:
It does exactly what it says. We now wait for the gear script manager to tell us it has given us all the gear we need, and then wait 325ms for all that gear/our entity to initialize.

This fixes the NVG issue we were having.
Also, we no longer need to set our radio channels through the client since the base game already does that, but we were just too fast in initializing the player.

Also, some standardization of Enums/Classes, while I was at it.

### DIFF
--- a/scripts/Game/Systems/Core/Managers/CRF_PlayerControllerManager.c
+++ b/scripts/Game/Systems/Core/Managers/CRF_PlayerControllerManager.c
@@ -56,7 +56,69 @@ class CRF_PlayerControllerManager : ScriptComponent
 	}
 
 	//------------------------------------------------------------------------------------------------
-	// INITIALIZATION
+	// INITIALIZATION AUTHORITY
+	//------------------------------------------------------------------------------------------------
+	
+	/**
+	 * (ONLY ON THE AUTHORITY) Initializes a looped check if the local player character has all gear from the GS manager
+	 * @param playerCharacter - Player character to pass along to InitilizePlayerCharacter from the gamemode manager
+	 * @param faction - Faction to pass alongto InitilizePlayerCharacter from the gamemode manager
+	 */
+	void InitilizePlayerCharacterCheck(SCR_ChimeraCharacter playerCharacter, Faction faction)
+	{
+		// Get the playable character comp from our entity so we can check if the gearscript has completed
+		CRF_PlayableCharacter playable = CRF_PlayableCharacter.Cast(playerCharacter.FindComponent(CRF_PlayableCharacter));
+		
+		// Check if we are already in a loop check, if we arent, start a loop check
+		GetGame().GetCallqueue().CallLater(PlayerCharacterCheck, 50, true, playable, playerCharacter, faction);
+	};
+	
+	/**
+	 * (ONLY ON THE AUTHORITY) Loop that checks if the local player character has all gear from the GS manager
+	 * @param playerCharacter - Player character to pass along to InitilizePlayerCharacter from the gamemode manager
+	 * @param faction - Faction to pass along to InitilizePlayerCharacter from the gamemode manager
+	 */
+	protected void PlayerCharacterCheck(CRF_PlayableCharacter playable, SCR_ChimeraCharacter playerCharacter, Faction faction)
+	{
+		// If gearscript hasnt been completed, dont initilize the entity
+		if (!playable.GetGearscriptCompleted())
+			return;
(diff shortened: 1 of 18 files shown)


scripts/Game/Systems/Core/Managers/CRF_PlayerControllerManager.c
```diff
@@ -56,7 +56,69 @@ class CRF_PlayerControllerManager : ScriptComponent
...
+	 * @param playerCharacter - Player character to pass along to InitilizePlayerCharacter from the gamemode manager
+	 * @param faction - Faction to pass alongto InitilizePlayerCharacter from the gamemode manager
+	 */
+	void InitilizePlayerCharacterCheck(SCR_ChimeraCharacter playerCharacter, Faction faction)
+	{
+		// Get the playable character comp from our entity so we can check if the gearscript has completed
+		CRF_PlayableCharacter playable = CRF_PlayableCharacter.Cast(playerCharacter.FindComponent(CRF_PlayableCharacter));
+		
+		// Check if we are already in a loop check, if we arent, start a loop check
+		GetGame().GetCallqueue().CallLater(PlayerCharacterCheck, 50, true, playable, playerCharacter, faction);
```
tanaaKa, line 73: Isnt 50ms a little too fast to be spawning threads and deleting them on all clients at once? As an aside, I feel as thoguh there's gotta be a better way than spawning and deleting threads over and over again to do a loop.

scripts/Game/Systems/UI/Menus/CRF_CharacterLoading.c
```diff
@@ -0,0 +1,42 @@
+class CRF_CharacterLoading: ChimeraMenuBase
+{
+	protected bool m_bFailsafeActive;
+	
+	/**
+	 * Called when the menu is opened
+	 * Initializes UI elements and sets up event listeners
+	 */
+	override void OnMenuOpen()
+	{
```
tanaaKa, line 9: Shouldn't we super these?

scripts/Game/Systems/UI/Menus/CRF_CharacterLoading.c
```diff
@@ -0,0 +1,42 @@
...
+	override void OnMenuOpen()
+	{	
+		GetGame().GetCallqueue().CallLater(ActivateFailsafe, 5000, false);
+	}
+	
+	/**
+	 * Cleans up resources when the menu is closed
+	 * Removes event listeners and slot update callback
+	 */
+	override void OnMenuClose()
```
tanaaKa, line 18: Shouldn't we super these?

tanaaKa:
Reviewed code and gameplay functionality. No explicit errors in code or game play and everything seems functioning.

scripts/Game/Systems/Core/Managers/CRF_PlayerControllerManager.c
Njpatman: I mean, we aren't; these are all on the server; they don't touch the clients at all. And we aren't "spawning threads and deleting them" it's just one thread that's constantly looping till the conditions are met to delete the thread and initialize the player.

scripts/Game/Systems/UI/Menus/CRF_CharacterLoading.c
Njpatman: Yeah, am dumb

scripts/Game/Systems/UI/Menus/CRF_CharacterLoading.c
Njpatman: Yeah, am dumb

scripts/Game/Systems/Core/Managers/CRF_PlayerControllerManager.c
tanaaKa: That's not what callLater() does. callLater spawns a thread every time it's called. This means when you specify 50ms thread spawns and then keep cancelling the thread, it is very taxing. The alternative would be to have a bool of some sort and have it done on frame or timeslice though.

Either way, we spawn threads A LOT and it's not good practice to spawn a thread in order to check a value and cancel it again.

tanaaKa: Future enhancement to clean those up imo

PR #625, opened 2025-05-24
Fix not being able close pause menu while respawn screen is open

MrPvTDagger:


Njpatman:
Just do call instead of call llater if you have the delay time to 0

Njpatman (approved):

PR #636, opened 2025-05-26
Alt character init method

Njpatman:
* Instead of creating multiple threads for each client controller, have it spawn threads based on the main thread.
* Re-add the radio initialization since we need that for respawns to get the correct radio channel.


Njpatman:
@tanaaKa ready for review

tanaaKa (approved):
Verified working in dedi with spawning and respawning

PR #734, opened 2025-08-10
Spawn spectators at point of death

tanaaKa:
Title


tanaaKa:
Abandoned since this change is included in https://github.com/CoalitionArma/Coalition-Reforger-Framework/pull/736

PR #735, opened 2025-08-10
Disable killfeed for admins

tanaaKa:
Title. Really stupid system tbh

Disables killfeed for admins NOT in zeus. When in zeus, killfeed is still visible.


tanaaKa:
Abandoned since this change is included in https://github.com/CoalitionArma/Coalition-Reforger-Framework/pull/736

PR #737, opened 2025-08-10
Register new gearscript configs

MrPvTDagger:
Regenerates config for admin menu when a new gearscript config is created


tanaaKa:
Can you include a readme into this or alter the existing readme with instructions on how to use this plugin?

scripts/WorkbenchGame/CRF_GenerateGearScriptConfigList.c
```diff
@@ -38,7 +38,16 @@ class AutoGenerateGearIndexPlugin : WorkbenchPlugin
...
 		ctx.SaveToFile("configs/GearScripts/GearScriptsConfigList.json");
 
-		Print("Saved to Configs/GearScripts/Standard/GearScriptsConfigList.json");
+		Print("Saved to Configs/GearScripts/GearScriptsConfigList.json");
+	}
+	
+	// This method is executed every time some new resource is registered
+	override void OnRegisterResource(string absFileName, BaseContainer metaFile)
+	{
+		if (!absFileName.Contains("CRF_GS_"))
```
Njpatman: Can you add even more filtering, so that it only runs if the file is in "Configs/GearScripts/"? 

I don't like the fact that anything with CRF_GS_ (Characters) can cause the configs to be reset.

PR #773, opened 2025-09-01
Players Can Now Toggle Satchels Between Throwable/Non-Throwable

### LINKED ISSUE #585: [CRF Feature] Add Interaction To Satchels To Toggle Between Throwable/Placable

Njpatman:
Fixes: #585


tanaaKa (approved):
I think this is fine, but best case scenario imo is the satchel is throwable by default.

RandAlThor1:
we could just swap the satchels in the gearscripts to the throwable variants by default

PR #778, opened 2025-09-02
Coalition von port

Salamiboio:
This Pull Request will:
Add CVON to the CRF as well as the mission below as a test/example mission for CVON.

========================================CVON======================================

99% of the time nothing has to be done on the mission makers end. For example the only change I had to make in my mission was making my vehicle groups belong to the platoons and not the company.

Any sort of vehicle element, armor, trucks, THs, etc... will be assigned to the highest headquarters at default which is the top most group. So if you ever want that changed you will have to override that. Same for multiple MMGs in one platoon as MMG1-3 go to Platoons 1-3. If theres only one MMG it just defaults to 1PLT, same with all other support groups like that.

I don't know how we want to approach this, if we just want to do this mission on Friday as a CVON test or during the EU CFC. We can keep this branch separate in case some game breaking bug comes up.

========================================Mission=====================================
Add the mission Ruha Salient made by Salami to the repository.

- [X] Mission File
- [X] Gearscripts Tested
- [X] Ingame Play Area Markers
- [X] Ingame Briefings to all players
- [X] Side-based briefings  
- [X] Mission .conf file

**Faction(s):**

BLUFOR - 80s Royal Marines
BLUFOR Comp: Motorized
BLUFOR Assets: 
1 Gun Truck per squad.

OPFOR - 80s USSR
OPFOR Comp: Motorized
OPFOR Assets: 
1 BTR Per Squad
1 MMG

BLUFOR: Secure all 3 RUSH points before mission end.
OPFOR: Ensure C does not get captured before mission end.

**Terrain:**

Ruha

A British company using motorized transport assault a contested airfield and the surrounding area again Soviet occupying forces.

**Slotting Ratio:**

2:1

**Time limit:**

70 Minutes

image

**Design concept/thoughts:**
Motorized combat over some of the more open terrain we have in the CRF.

**Other Notes:**

There's trucks spawned for all 3 respawn phases, does not mean take them. If a GT or BTR team is killed they can take the new one when they respawn, that's it.


scripts/Game/Systems/Core/CRF_PlayableCharacter.c
```diff
@@ -252,19 +252,10 @@ class CRF_PlayableCharacter : ScriptComponent
...
-				mat[0], mat[1], mat[2], mat[3]
-			);
-		}
+		mat[1] = vector.Up;
+		mat[2] = vector.Forward;
+		mat[3][1] = 10000;
 		
-		mat[3][1] = mat[3][1] - 1.5;
 		m_PlayerControllerComponent.UpdateEntityPos(mat);
 	}
```
Njpatman: Why are we throwing out how specs save their last known position for a VON import?

Salamiboio: They no longer need to know it, they hear from the camera now not the entity.

Salamiboio: They no longer need to know it, they hear from the camera now not the entity.

Prefabs/Characters/CRF_InitialEntity.et
```diff
@@ -176,7 +176,6 @@ SCR_ChimeraCharacter : "{37578B1666981FCE}Prefabs/Characters/Core/Character_Base
   BaseLoadoutManagerComponent "{520EA1D2DB11821E}" {
    Slots {
     LoadoutSlotInfo Back {
-     Prefab "{13A97D10A827AE01}Prefabs/Items/Equipment/Radios/SpecRadioBag.et"
```
Njpatman: Why?

Salamiboio: We don't use radios to talk anymore, just direct.

Njpatman: so, it'll be like everyone is just talking like it's a standard TS/Discord channel?

Salamiboio: Yeah, in spectator, rooms already work with it as well.

tanaaKa:
Needs a master on/off switch to avoid becoming a dependency that is required, for now.

Salamiboio:
I can add a turn on and off switch however comma we'd have to have two separate GS Global Role Configs and someway to switch between the CVON radios and Regular Radios. Unless you mean like give the mission the choice to use it and we just have to chose one system over the other.

tanaaKa:
Everything you're proposing would entail a complete overwrite as is in this PR. This PR needs to introduce a way for us to use both in-game VON, as well as CVON if the MM says that. We shouldnt plan to overwrite functionality, at least for now, so this PR will need to introduce a way for both systems to exist.

Salamiboio:
For now I'll create the two separate global configs then and two CRF_Lobbys. One that supports CVON and one that doesn't. All the MM does is has to put which one in the world and it'll work with whatever system from there. Sound good?

tanaaKa:
Works for me

Salamiboio:
Well fuck totally just realize I have to rip up all the spectator work I did :/

tanaaKa:
Well, not trying to be the bearer of bad news, but we can't simply force this on everyone yet. Likewise, we shouldnt have multiple branches which will create excess overhead for us developing things. The best case scenario is to separate the logic in-mod that allows us to use both.

Salamiboio:
Fair, I'll focus on this tomorrow, this is a pretty monumental task as we have base game UI Overrides as well that I have to make work with vanilla now

Salamiboio:
I'll post the readme are we good to do a test thursday for this?

tanaaKa:
Yep

Salamiboio:
You wanna announce it or me?

Salamiboio:
All the MM has to do to disable it now is to just disable the gamemode component. That's it. The CVON Radis function just like the COALITION Radios now with their channels so the gearscript prefabs don't need to be changed.

Salamiboio:
Only other change I could see is throwing a download up for the plugin on the coalition website and directly linking to that instead of the discord

tanaaKa:
@Salamiboio Last call on this PR. Please clean up anything that you've changed or have discovered and resolve the merge conflicts.

Mission makers will be using this this week.

Salamiboio:
I'm on guard shift for the next 9 hours so I won't be able to double check this, should be good for now all current reported bugs are on the CVON side.

tanaaKa:
Ok I'm going to add a mission maker and end user guide.

https://github.com/CoalitionArma/Coalition-Reforger-Framework/blob/790c8298ddf570d557842c40c5d2ea9f744ecf58/docs/CVON_MISSION_SETUP_TUTORIAL.md

PR #780, opened 2025-09-03
Remove Dragger

Salamiboio:
Removes the Dragger mod as we have ACE Draggin


Salamiboio:
Actually looking deeper ace carry on has carrying strictly, this did have a bug happen to Outtake last CFC where she was dragging someone and couldn't release them. I had to heal the person she was dragging to have him be released.

PR #785, opened 2025-09-06
Add CRF_VehicleDepot.c

TristGit:
Vehicle Depot component meant to be attached to objects to act as the hub/depot for spawning. It integrates nearby supplies, can utilize faction tickets, or a custom 'Use' attribute- Or all of them at once, per vehicle. Customizable positions, spawn pattern, and spawn collision detection (so no multiple vehicles spawning on same spawn, killing players, or intersecting with objects.).
image

- (Should be) 2 Plug and play prefabs for MM's outside of adding new entries into Actionsmanager for additional vehicles and customizations. Prefab/Entity bound scripting only.
- Spawn pattern types
- 3 possible resources: tickets, uses, and/or supplies (registers -1 for tickets OR -1 for uses as unlimited)
- Detects closest supply box/vehicle with supplies within set radius (switches to next after done), can be unloaded into of prefab with cache
- All depots have independent custom Use pools (though all share settings for the pool usage amount). 
- Having too many depots in-scene may/will flood network with update calls, via it's cache/resource/menu update loop. Did my best to mitigate. Prejoin would be good resting ground.

**Vehicle Depot System Implementation**

* Added new prefabs `VehicleDepot.et` and `VehicleDepot_Cache.et` with configuration for vehicle arrays, spawn actions, mesh objects, and resource containers, supporting multiple cost types (tickets, supplies, uses) and spawn patterns. 
* Created corresponding metadata files for both prefabs to enable platform-specific configurations. 

**Scripting and Server Support**

* Created a `CRF_DepotSpawnAction.c` subclass meant for the ActionsManagerComponent.c, handling the menu creation for spawning vehicles, including cost checks, error handling, and optimized display text updates. Supports both direct spawn and dedicated server RPC via `CRF_RplToAuthorityManager`.
* Extended `CRF_RplToAuthorityManager.c` with new RPC methods to handle vehicle spawn requests on the server, ensuring compatibility with dedicated servers and proper authority management. 

**Documentation**

* Added `VEHICLE_DEPOT_USER_GUIDE.md` detailing setup steps, configuration options, cost types, troubleshooting, and custom depot creation for mission makers and users.

**Personal Notes**

- I implemented maybe too many optimizations via rate limiting or cacheing, but testing via server and/or client seems to work when configured correctly. Check loops are fired either only when the action menu radius is entered (vanilla CRF_ActionsMenuManager), after spawning an object (updates caches for dynamic resource checks text display for the actions), and limited RPC to spawning/resource deduction. Only perpetual loop is checking for faction ticket changes every 15 seconds.
- I have an entry within RplToAuth (vehicle spawning only, had to be called in order for actionsmanager to properly fire) because it just felt okay considering but let me know if the implementation is bunk or for more expansive systems.
- Let me know of any other ideas or changes or anything else. Doc is included. This is ALMOST too old of a component as its sat unpublished on my local for a while and I struggled to properly handle replication the last few days
- Heavily commented / some rolling changes were made via copilot. Inform if unsuitable

Needs proper testing by someone who isn't busy and/or sleep deprived


TristGit:
fyi, supply consumption broken/stuck onit atm. I dont know why or how it broke after testing, but now its bunk. Will eventually fix.

PR #793, opened 2025-09-12
RPC Batch Calls

tanaaKa:
This PR changes a majority of the major RPC calls to be batched in that the call uses interior checking to be picky and see what - if anything - needs to be replicated in order to avoid unnecessary overhead.

According to LLM Claude:
- Approx 95% reduction in rpc/network traffic
- 1 RPC per slotting update now

Still pending:
- Slot data container optimization to come later

LLM overview:
### Batched Slot Update System
- **Replaced multiple individual RPCs with a single batched RPC** for slot updates
- **Added `CRF_SlotUpdateData` struct** to batch all slot properties into one network message
- **Added `BatchUpdateSlot()` method** for efficient slot updates
- **Modified existing slot update methods** to use batched updates while maintaining backward compatibility

### Before Optimization
- **6-7 individual RPCs** sent for each slot change:
  - `RpcAsk_UpdateSlotPlayerID`
  - `RpcAsk_UpdateSlotLockedState`
  - `RpcAsk_UpdateSlotDeathState`
  - `RpcAsk_UpdateSlotGroup`
  - `RpcAsk_UpdateSlotResource`
  - `RpcAsk_UpdateSlotCharacter`
  - Plus additional calls through `RequestSlottingUpdate()`

### After Optimization
- **1 single RPC** for all slot changes:
  - `RpcAsk_BatchUpdateSlot` handles all properties at once
- **Conditional updates**: Only updates properties that actually changed
- **Backward compatibility**: Existing API still works but uses optimized system

### Network Traffic Reduction
- **85-95% reduction** in slot-related network messages
- **Before**: Up to 280+ RPCs for 40 slots with full updates
- **After**: Maximum 40 RPCs for 40 slots (one per slot)

### Late Joiner Performance
- **Dramatically reduced data synchronization** when players join mid-game
- **Fewer replication calls** means less processing overhead
- **Batched updates** reduce network congestion

### New Components Added

1. **Simplified Batched RPC Pattern**
   ```c
   // Client-side call
   void BatchUpdateSlot(int slotId, int playerId, RplId groupId, RplId charId, 
                       ResourceName resource, string name, bool isLocked, bool isDead)
   {
       Rpc(RpcAsk_BatchUpdateSlot, slotId, playerId, groupId, charId, resource, name, isLocked, isDead);
   }

   // Server-side handler  
   [RplRpc(RplChannel.Reliable, RplRcver.Server)]
   protected void RpcAsk_BatchUpdateSlot(int slotId, int playerId, RplId groupId, RplId charId, 
                                        ResourceName resource, string name, bool isLocked, bool isDead)
   ```
2. **Single RPC call with all slot properties**
   - Replaces 6-7 individual RPCs with 1 batched RPC
   - Uses simple parameter passing (no complex data structures)
   - Follows PlayableSelector reference implementation pattern
3. **Server-side batching and conditional updates**
   - Only applies changes to properties that actually differ
   - Single `RequestSlottingUpdate()` call instead of multiple
   - Validates slot data exists before attempting updates
### Backward Compatibility
- All existing `UpdateSlot*()` methods still work
- Methods now attempt to use batched updates when possible
- Falls back to individual RPCs if slot data unavailable
- No breaking changes to existing calling code
### Old Way (Still Works)
```c
UpdateSlotPlayerID(slotId, playerId);
UpdateSlotLockedState(slotId, true);
UpdateSlotDeathState(slotId, false);
// Results in 3 separate RPCs
```

### New Optimized Way
```c
BatchUpdateSlot(slotId, playerId, groupId, charId, resource, name, true, false);
// Results in 1 single RPC
```
### Phase 1: Immediate (COMPLETE)
- ✅ Batched update system implemented
- ✅ Backward compatibility maintained
- ✅ Automatic optimization for existing calls
### Phase 2: Code Updates (RECOMMENDED)
- Update calling code to use `BatchUpdateSlot()` directly when multiple properties change
- Remove redundant individual update calls in sequence
- Monitor performance improvements
### Phase 3: Cleanup (FUTURE)
- Once all calling code updated, can remove individual RPC handlers
- Further optimize by removing fallback mechanisms
## Next Steps for Additional Optimization
1. **Review CRF_SlottingManager.RequestSlottingUpdate()**
   - This method still calls `Replication.BumpMe()` which forces full replication
   - Consider implementing proper RplProp usage instead
2. **Implement RplSave/RplLoad for static slot data**
   - Separate load-time from runtime data
   - Static slot configuration should use RplSave/RplLoad
3. **Convert to individual slot entities**
   - Consider implementing Solution 1 for even better performance
   - Individual slot entities would eliminate the need for array synchronization
## Files Modified
- `scripts/Game/Systems/Core/Managers/CRF_RplToAuthorityManager.c`


tanaaKa:
Closing since new PR includes random initial entity spawning and I didnt want to rebase
https://github.com/CoalitionArma/Coalition-Reforger-Framework/pull/794

PR #799, opened 2025-09-15
Updated to support Game Master/Spectator

Salamiboio:
Just makes audio listening come from the camera now instead of the player character


tanaaKa:
@Salamiboio Handle the conflict first please

Salamiboio:
Resolved

PR #826, opened 2025-09-25
Ticket tracking refactor

### LINKED ISSUE #782: [CRF General Bug] Subtracting tickets is bugged
void SubtractTicket(FactionKey faction, int amount, bool force = false)
    {
        // Don't subtract tickets during safestart
        if (m_SafestartManager.GetSafestartStatus() && !force)
            return;
            
        int currentTickets = GetFactionTickets(faction);
        
        // Only subtract if tickets are positive and not unlimited (-1)
        if (currentTickets <= 0 || currentTickets == -1)
            return;
            
        // Update the appropriate faction's tickets
        switch (faction)
        {
            case "BLUFOR": m_Gamemode.m_iBLUFORTickets -= amount; break;
            case "OPFOR": m_Gamemode.m_iOPFORTickets -= amount; break;
            case "INDFOR": m_Gamemode.m_iINDFORTickets -= amount; break;
            case "CIV": m_Gamemode.m_iCIVTickets -= amount; break;
        }
        
        Replication.BumpMe();
    }

Two issues, we are using an attribute variable to track current tickets which should not happen. Second we are calling Replication.BumpMe(); outside of the object that we should be updating. We need to move all the tickets to the respawn manager and make them RplProps.

tanaaKa:
Fixes #782. @Salamiboio please review.

### DIFF
--- a/scripts/Game/Systems/Components/VehicleDepot/CRF_VehicleDepot.c
+++ b/scripts/Game/Systems/Components/VehicleDepot/CRF_VehicleDepot.c
@@ -432,20 +432,11 @@ class CRF_VehicleDepot : ScriptComponent
 		FactionKey factionKey = playerFaction.GetFactionKey();
 		
 		// Get tickets from CRF_Gamemode based on faction
-		CRF_Gamemode gamemode = CRF_Gamemode.GetInstance();
-		if (!gamemode)
+		CRF_RespawnManager respawnManager = CRF_RespawnManager.GetInstance();
+		if (!respawnManager)
 			return false;
 			
-		int currentTickets = 0;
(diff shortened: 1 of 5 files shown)


Salamiboio:
So now if there is no tickets it's just -1 and anything above that is a side has tickets?

tanaaKa:
The logic hasn't changed at all, nor should it. This is a PR to fix the subtracting bug. If we want to talk logic, we should create a new issue.

Salamiboio:
Ohhhhh my bad, I see now, tested it and it looks good.

Salamiboio (approved):

PR #827, opened 2025-09-25
Add admin camera to mods

### LINKED ISSUE #748: [CRF Feature] Allow mods to use admin vision spectate cam
There a admin vision like the arma vision would be useful for the mods to use when helping with tickets so we can see what going on

tanaaKa:
This should work and will close #748. Please verify @MrPvTDagger.


MrPvTDagger (approved):
Verified in dedicated server tools

PR #846, opened 2025-09-27
Radio Init Fix

### LINKED ISSUE #843: [CRF General Bug] Radios still aren't initializing on gear reset
Just gotta increase the delay times as the client is not reinitializing

Salamiboio:
Fixes https://github.com/CoalitionArma/Coalition-Reforger-Framework/issues/843


Njpatman (approved):
I dont like just adding more delay, but if it works, it works

PR #853, opened 2025-09-30
CRF_VehicleDepot optimization

TristGit:
Optimizations to steer away from radius detections + callLaters to handle supply detection / aggregation, it's a temporary fix due to me throttling the GetActionNamescript which fires on every frame to ONLY refresh supplies server side once per 10s / 12s clientside (to prevent rpc's). More optimized, though limiting a base Reforger function that fires per frame isnt ideal in the long run,

 I want to eventually have it handled via its own menu instead to get past menuaction limitation.


Njpatman:
Please delete/comment out any prints

Njpatman (approved):

PR #876, opened 2025-10-05
GS NVG Fixes

Salamiboio:
Fixes NVGs in the GS


Njpatman (approved):
May we pray the call queue doesn't explode

PR #878, opened 2025-10-05
Tweaks shaped charge changes.

HarryEngineering:
Reduces deadly fragmentation cone, adds more light spalling to hopefully reduce weapon efficiency.

Should note that this is only affecting my _LS prefabs, not all warheads.


Njpatman:
Have you done testing? Will crewmembers be killed easily, will vehicles be harder to destroy, etc?

HarryEngineering:
Yes, I've tested. Vehicle destruction wasn't the problem, it was the efficient crew kill. This change is mainly aimed at hedp, but also tweaks some penetrator changes I think would be too much given the experience with it. 

You can see in the image below a recreation of the gaz crew cabin. Testing also reinforced just how perfect the shot was, because It's very difficult to get more than 1 kill with either version, and either of my tweaks are less generally deadly than the shrapnel mod made hedp. 

image

And what the fragmentation pattern looks like naturally. 

image

HarryEngineering:
Actually on retest, I may up the spread a touch to try to wound more. I think this is too much the other way.

tanaaKa:
Fwiw, I'm 50/50 on any changes, but it's good to have options. It makes most sense to use HE for 40mm only as grenadiers should never truly be engaging vehicles.

HarryEngineering:
I'm fine with w/e. Even if we discard HEDP, fragmentation knowledge and changes will still be useful for AT launchers which are included in the testing.

PR #896, opened 2025-10-07
Cvon fixes

Salamiboio:
Fixed radio frequencies not being assigned after mini arsenal update.
Fixed radio frequencies sometimes not assigning when loading in.
Fixed radio settings not applying on respawn or gearscript reset.
Fixed radios from appearing in RHS radio pouches.
Fixed radio icon sizes.


Salamiboio:
Sorry for multiple fixes in 1 PR, most of these I couldn't fix without the other being fixed as well. Should resolve most of our radio issues.

I'll state this here cause there's an overall problem with the Gearscript in its implementation as it stands. We need to add a query system to it or something so we do not have to rely on CallLaters. 

The reason we were having these issues was a bug with the game code where assigning an item to an inventory immediately after creation would occasionally cause the EntityEvents flags to not be set on the server for some reason. I worry this will break more than just my radios, I've fixed it for now but changing the implementation of how we initially init a radio but I worry this will break other mods in time.

A query system will allow us to add items to inventories after set intervals we set up so we do not have to rely on the callqueue and it potentially exploding the Replication thread.

Njpatman (approved):

PR #905, opened 2025-10-09
Entity Init Refactor

### LINKED ISSUE #903: [CRF General Bug] Player Init fails on relog
Whenever you relog or occasionally on first join you don't get thrown into spectator or the UI.

Salamiboio:
Fixes https://github.com/CoalitionArma/Coalition-Reforger-Framework/issues/903


Salamiboio:
Can't 100% test reconnecting with an existing slot as the dedicated server tools in 1.4 generates a brand new player id every time you open the diag exe, but it seems to be working.

Also has some minor fixes like the spectator camera being in the stratosphere on death and never going to the camera position where you died at.

Much cleaner now as the only places we request entity init is now on the server and the only time the client requests it is when you go in game to refresh your entity.

Njpatman (approved):
let's see what happens

Issue #776, opened 2025-09-02
[CRF Feature] Add Gearscript support to vehicles.


Salamiboio:
What I mean is like in our old A3 system where we can define certain vehicles to get supplies in the inventory based off of that gear script. Makes logi trucks easier to place and ensures they have the equipment for their element.

Salamiboio:
Ties into https://github.com/CoalitionArma/Coalition-Reforger-Framework/issues/781 as we'll populate a vehicles inventory when it spawns in.

Njpatman:
Please ensure you stick to the CRF_GS prefix for any vehicles, especially if you're gonna be adding custom components to these vehicles (and ensure you add the vehicle to each GS faction and their respective Zeus editable entity lists)

Salamiboio:
Why would we want to limit what vehicles we can add loadouts to? This is like the old arma 3 system where unless it was a logi truck, it'd have an extra loadouts worth of supplies. If it is a logi truck prefab, then it gets the full logi dump.

RandAlThor1:
hear me out. in the vehicle spawner you are making. we add an option to select "type" so that it will auto draw from the gearscript for its inventory as if it was the type: (truck, tank, IFV, ect ect)

Salamiboio:
My current plan is where we define gearscript in the gamemode well also define Logistic Trucks for the mission where we can have our defaults CRF_GS of course. Then every other vehicle that gets spawned gets the default loadout. However, comma, you do bring up a good idea. Whenever a vehicle is spawned, we can set up its ammo, so it's always right, like 400 rounds HE and AP instead of the issue where it's a MM oversight and the BMP only has 100 rounds of coax.

RandAlThor1:
IMO, we should just have preset configs for inventories similar to gearscripts it can be a separate config for a sake of ease of use. we can have preset ZEUS vehicles for the sake of spawning mid mission. as well as if we can make use of that fact that EVERY vehicle in a mission should be being spawned via he spawner to control the pre-placed ones. 

zeus ones would just draw from the gearscript as well

Salamiboio:
I was thinking the same, we can have some global config that details the amount of ammo, let's say calibers of weapons have cause you can pull that info and it populates off that, AP and HE canons are seperate rn for autocanons so this works out. We can also just override SCR_Vehicle constructor or OnPostInit so it works for all vehicles spawned or GM placed for an admin fix.

Salamiboio:
[...]

PR #911, opened 2025-10-10
Vehicle gear script

Salamiboio:
Adds https://github.com/CoalitionArma/Coalition-Reforger-Framework/issues/776
Current Default Ammo:
image
Default Loadout:
300 AP
300 HE
1200 MG

You can override base weapon ammo values, vehicle loadouts and additional inventory items faction wide and per vehicle as well so it's never locked and the MM doesn't need to touch these global values if he want's to change some micro ammo assignment.


Salamiboio:
In the future if we add back tanks I'll have to add them to the GS as I obv don't have access to a Canon weapon type right now.

Salamiboio:
Can someone review this

tanaaKa:
Resolve conflicts and will merge

Salamiboio:
Fixed

PR #967, opened 2025-10-27
UGL/Demo/Medic  Update

RandAlThor1:
Changes

- Removes SL/TL UGL from global configs
- Replaces 1 AT and the Demo from the default squad to 2x Gren
- Re-adds squad medics

Total 2x UGLs in default squad


tanaaKa:
My 2c, lets keep this 2x RAT. People have already complained of 40mm spam and this allows for default squads to still be standard for combined operations. SLs should still get a UGL as well. 2x RAT, 2x 40mm in squad.

HarryEngineering:
Can we close this one way or the other so people can balance their missions coming up? A suggestion is to add an SL and TL UGL box to the GS like we have for medic frags, and default to SL on with two RAT.

Salamiboio:
Cant review it grrr

Salamiboio:
Make a seperate feature request for the SL and TL GL gearscript feature

PR #1002, opened 2025-11-02
Feature - Adds Vehicle Rearming to Ammo Trucks

Salamiboio:
Overhauls the ammo truck

- Gives it a menu to rearm local vehicles, this just reinitializes their gear script
- Increase supply count from 450 to 1500 so they can actually rearm vehicles


Salamiboio:
Please make sure this is reviewed before company session as I want to use it there

Njpatman (approved):

PR #1027, opened 2025-11-04
Feature - Disables AI Crouching

Salamiboio:
Adds a setting in the gamemode to disable or enable ai crouching to combat them shooting through the tall grass while prone.


tanaaKa:
I'm sure we're going to add more AI customization as we run into it. Create a new divider on the game mode for this. Something along the lines of "CRF AI" etc...

PR #1036, opened 2025-11-07
Reworks the ticket replication

MrPvTDagger:
Moved the ticket replication to server authoritative. Should be lightweight network wise as the array is small and is only called from the server when a admin has the menu open.

What this fixes
- Tickets no longer tied to the player being connected (if a player disconnected it wouldn't let you reply, see message history or close the ticket)
- If a admin disconnects or jips they will get back all the tickets that are currently open

Small Fixes
- Chat now updates when admin menu is open
- Add more feedback for when a ticket is closed or a admin is assigned to a ticket for other admins to see


scripts/Game/Systems/Core/Managers/CRF_AdminMenuManager.c
```diff
@@ -5,6 +5,58 @@ class CRF_TicketMessageData
...
+	{
+		snapshot.DecodeString(packet);
+		snapshot.DecodeString(packet);
+		snapshot.DecodeString(packet);
+		
+		return true;
+	}
+	
+	static bool SnapCompare(SSnapSerializerBase lhs, SSnapSerializerBase rhs, ScriptCtx ctx)
+	{
```
tanaaKa, line 48: Why is this doing the same thing three times? IIRC, should be:

```
static bool SnapCompare(SSnapSerializerBase lhs, SSnapSerializerBase rhs, ScriptCtx ctx)
{
    if (!lhs.CompareStringSnapshots(rhs)) return false;  // sender
    if (!lhs.CompareStringSnapshots(rhs)) return false;  // msg
    if (!lhs.CompareStringSnapshots(rhs)) return false;  // timestamp
    return true;
}
```

scripts/Game/Systems/Core/Managers/CRF_RplBroadcastManager.c
```diff
@@ -59,15 +59,32 @@ class CRF_RplBroadcastManager : ScriptComponent
 	}
 	
 	//------------------------------------------------------------------------------------------------
-	void SendAdminMessage(string data, int playerID)
+	void SendAdminMessage(string data, int playerID, bool ticketExists)
```
tanaaKa, line 296: There's a race condition issue here where `ticketExists` is calculated on the client, so if the client is slow, it's entirely possible this gets sent as false.

Recommend:
```
[RplRpc(RplChannel.Reliable, RplRcver.Server)]
protected void RpcAsk_SendAdminMessage(string data, int playerID)
{
    bool ticketExists = m_AdminMenuManager.TicketExists(playerID);
    m_RplBroadcastManager.SendAdminMessage(data, playerID, ticketExists);
    // ... rest of the logic
}
```

scripts/Game/Systems/Core/Managers/CRF_AdminMenuManager.c
```diff
@@ -129,7 +182,33 @@ class CRF_AdminMenuManager : ScriptComponent
...
-		RefreshLists();
+		CRF_RplBroadcastManager.GetInstance().RefreshAdminMenuLists();
+	}
+		
+	/**
+	* Keeps track of the admin that helping with this ticket
+	* @param ticketID ID of the ticket (ID of the player that initially opened the ticket)
+	* @param admin ID of the admin that will be helpign the player
+	*/
+	void AssignAdminTicket(int ticketID, int adminID)
```
tanaaKa, line 194: There's a data flow issue here where this is called on the server, modifies local state, then broadcasts. But the RpcDo_AssignAdminTicket handler also runs on the server due to RplRcver.Broadcast, potentially causing double-processing.

Add something like 
```
(!Replication.IsServer())
```

scripts/Game/Systems/Core/Managers/CRF_RplBroadcastManager.c
```diff
@@ -564,30 +630,48 @@ class CRF_RplBroadcastManager : ScriptComponent
...
+		if (!chatComponent)
+			return;
 
 		// Display a admin assigned them self to the ticket
-		m_AdminMenuManager.NewTicketMessage(ticketID, adminID, "Assigned to Ticket");
+		chatComponent.ShowMessage(string.Format("%1 assigned to %2's ticket", adminName, playerName));
+	}
+	//------------------------------------------------------------------------------------------------
+	[RplRpc(RplChannel.Reliable, RplRcver.Broadcast)]
+	void RpcDo_RefreshAdminMenuLists()
```
tanaaKa, line 1229: This goes to all clients. You should add a check to ensure it's only be processed or going to admins.

tanaaKa (changes requested):

tanaaKa:
image

scripts/Game/Systems/Core/Managers/CRF_AdminMenuManager.c
MrPvTDagger: RplRcver.Broadcast doesn't run on the server

tanaaKa: You right. Would still be good imo.

MrPvTDagger: The code run in the broadcast isn't the same as the code in that method, bad naming. so pointless imo

MrPvTDagger: I've renamed it so less confusing

MrPvTDagger:
Would have thought it would re requested review when I close the reviews...

PR #1051, opened 2025-11-14
RPL optimization for slot data

tanaaKa:
Adds three things:

1. Dirty flag checking to verify we need to change anything at all 
2. Batching managed by-frame if we have too many network calls at once (loads them all onto the next frame)
3. Delta updates rather than widespread update


Njpatman (approved):
LGTM

PR #1077, opened 2025-11-24
Feature - Spawns spectators around the AO Center and Radius

### LINKED ISSUE #1076: [CRF Feature] Center initial entity spawns around AO not map center.

Salamiboio:
Fixes https://github.com/CoalitionArma/Coalition-Reforger-Framework/issues/1076

### DIFF
--- a/scripts/Game/Systems/Core/CRF_PlayableCharacter.c
+++ b/scripts/Game/Systems/Core/CRF_PlayableCharacter.c
@@ -17,7 +17,7 @@ class CRF_PlayableCharacter : ScriptComponent
 	protected IEntity m_eSpecEntity;
 	//So the client tracks where he needs to teleport his player
 	//Since teleporting is mostly client authorative
-	[RplProp()] vector m_vSpreadPos;
+	vector m_vSpreadPos;
 
 	//------------------------------------------------------------------------------------------------
 	override void OnPostInit(IEntity owner)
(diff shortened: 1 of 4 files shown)


scripts/Game/Systems/Core/CRF_PlayableCharacter.c
```diff
@@ -100,19 +102,32 @@ class CRF_PlayableCharacter : ScriptComponent
 		
 		if (isCRFInitialEntity)
 		{
-			vector mapSize = SCR_MapEntity.GetMapInstance().Size();
-			vector mapCenter = Vector(mapSize[0] / 2, 0, mapSize[1] / 2);
-			vector spreadPos = GenerateRandomSpreadPosition(mapCenter, 500.0);
-			spreadPos[1] = 1000.0; // Set elevation to 1000m
-			m_vSpreadPos = spreadPos;
-			owner.SetOrigin(spreadPos);
+			if (Replication.IsServer())
```
tanaaKa, line 105: DRY - Dont repeat yourself. Create a method to do this instead of doing the same thing twice.

scripts/Game/Systems/Core/CRF_PlayableCharacter.c
```diff
@@ -100,19 +102,32 @@ class CRF_PlayableCharacter : ScriptComponent
...
+			{
+				vector mapCenter;
+				float radius;
+				CRF_Gamemode.GetInstance().GetAOCenterAndRadius(mapCenter, radius);
+				vector spreadPos = GenerateRandomSpreadPosition(mapCenter, radius);
+				spreadPos[1] = 1000.0; // Set elevation to 1000m
+				m_vSpreadPos = spreadPos;
+				owner.SetOrigin(spreadPos);
+				Replication.BumpMe();
```
tanaaKa: I'd like to avoid `BumpMe()` if we can.

What do you think of:
```
class CRF_PlayableCharacter : ScriptComponent
{
    [RplProp()] vector m_vSpreadPos;
    
    // Server calculates and sends position to specific client
    [RplRpc(RplChannel.Reliable, RplRcver.Owner)]
    protected void RPC_SetSpreadPosition(vector pos)
    {
        m_vSpreadPos = pos;
        // Client receives and can act on it immediately
    }
    
    protected void SetRandomAOSpreadPosition(IEntity owner)
    {
        if (!Replication.IsServer())
            return;
            
        vector mapCenter;
        float radius;
        CRF_Gamemode.GetInstance().GetAOCenterAndRadius(mapCenter, radius);
        vector spreadPos = GenerateRandomSpreadPosition(mapCenter, radius);
        spreadPos[1] = 1000.0;
        m_vSpreadPos = spreadPos;
        owner.SetOrigin(spreadPos);
        
        // Send to owner client only
        Rpc(RPC_SetSpreadPosition, spreadPos);
    }
}
```

This way it only sends the specific data to specific clients. We don't need it in the global queue for all clients.

tanaaKa (changes requested):

scripts/Game/Systems/Core/CRF_PlayableCharacter.c
Salamiboio: It'd have to be requested from the client to the server and back but yeah that's probably the better option

Salamiboio:
Client now requests the generated pos for the spec entity so only the client who needs it gets sent the data

tanaaKa (approved):

PR #1087, opened 2025-12-05
Feature - Adds shareable markers

### LINKED ISSUE #1083: [CRF Feature] Disable global markers after safestart
Once safestart ends any marker placed afterwards should have to be shared to other players

Salamiboio:
Fixes https://github.com/CoalitionArma/Coalition-Reforger-Framework/issues/1083

Can be toggled off per faction by mission maker.

Default behavior is during safestart all markers are broadcasted like usual, after safestart players have to be within 8m of each other and share markers for them to be seen.


scripts/Game/Systems/VanillaOverrides/Managers/CRF_SCR_MapMarkerManagerComponent.c
```diff
@@ -1,6 +1,39 @@
...
+				marker.m_bIsShared = true;
+		
+		
+		super.OnAddSynchedMarker(marker);
+	}
+	
+	override void Update(float timeSlice)
+	{
+		super.Update(timeSlice);
+		foreach (SCR_MapMarkerBase marker: m_aStaticMarkers)
```
tanaaKa, line 53: foreach in a per frame update?

scripts/Game/Systems/VanillaOverrides/UI/CRF_SCR_MapMarkerBase.c
Njpatman, line 1: Hey, so why are we turning SCR_MapMarkerBase into a replicable class?

Salamiboio: It already is, any modded class which has these methods need them in the modded class override. Just a quirk of the IDE.

scripts/Game/Systems/VanillaOverrides/Managers/CRF_SCR_MapMarkerManagerComponent.c
Salamiboio: Other option is not doing the super and pluging into the foreach in the regular method.

Salamiboio: Then we run into issues with any other mod that overrides this class

tanaaKa:
I've pushed some changes. Review them if you'd like for functionality.

In short
- Moved away from a loop per frame 
- markers are only shared in same faction
- null checks
- m_bIsShared replication
- got rid of code deadends with breaks etc..

Salamiboio:
LGTM

Salamiboio:
- Removed m_bIsShared from serialization, not needed as it's a client sided value.
- Fixed bug with other methods in the marker manager setting the marker visible when it shouldn't be
- Fixed marker not being set to shared on a marker placed by the local client

Njpatman (approved):
LGTM, Gonna let @tanaaKa take a look as well if he wants

PR #1102, opened 2025-12-12
Feature - Reworks bushes

### LINKED ISSUE #1101: [CRF Feature] Bush rework
Should do 3 things

- Slow down movement speed when walking through a bush
- Prevent going prone in a bush
- Increase volume/distance of a player walking through a bush

Salamiboio:
Fixes https://github.com/CoalitionArma/Coalition-Reforger-Framework/issues/1101

### DIFF
--- a/scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
+++ b/scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
@@ -104,4 +104,10 @@ class CRF_BushMovementComponent: ScriptComponent
 		
 		m_DamageManager.SetMovementDamage(0.5);
 	}
+	
+	//Extra redundancy incase something fucking insane happens
+	void ~CRF_BushMovementComponent()
+	{
+		UnregisterEntity();
+	}
 }
--- a/scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
+++ b/scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
@@ -1,0 +1,107 @@
+class CRF_BushMovementComponentClass: ScriptComponentClass
+{
+}
+
+class CRF_BushMovementComponent: ScriptComponent
+{
+	[Attribute(params: "et")]
+	ref array<ResourceName> m_aBushPrefabs;
+	
+	float m_fOriginalSpeed = 0;
+	bool m_bEffectsApplied = false;
+	bool m_bEffectsAppliedThisFrame = false;
+	vector m_vOriginThisFrame;
+	protected SCR_CharacterControllerComponent m_CharacterController;
+	protected SCR_CharacterDamageManagerComponent m_DamageManager;
+	
+	void RegisterEntity()
+	{
+		//redundency to ensure this is NEVER on the server
+		if (Replication.IsServer())
+			return;
+		
+		m_CharacterController = SCR_CharacterControllerComponent.Cast(GetOwner().FindComponent(SCR_CharacterControllerComponent));
+		m_DamageManager = SCR_CharacterDamageManagerComponent.Cast(GetOwner().FindComponent(SCR_CharacterDamageManagerComponent));
+		SetEventMask(GetOwner(), EntityEvent.FRAME);
+	}
+	
+	void UnregisterEntity()
+	{
+		//redundency to ensure this is NEVER on the server
+		if (Replication.IsServer())
+			return;
+		
+		ClearEventMask(GetOwner(), EntityEvent.FRAME);
+	}
+	
+	float m_fBuffer = 0;
+	override void EOnFrame(IEntity owner, float timeSlice)
+	{
+		if (!m_DamageManager || !m_CharacterController)
+			return;
+		
+		if (m_fBuffer <= 0.1)
+		{
+			m_fBuffer += timeSlice;
+			return;
+		}
+		
+		m_fBuffer = 0;
+		
+		m_bEffectsAppliedThisFrame = false;
+		m_vOriginThisFrame = owner.GetOrigin();
+		m_vOriginThisFrame[1] = m_vOriginThisFrame[1] + 1;
+		GetGame().GetWorld().QueryEntitiesBySphere(m_vOriginThisFrame, 1, BushCheckCallback, null);
+		if (!m_bEffectsAppliedThisFrame && m_bEffectsApplied)
+			ResetBushEffects();
+	}
+	
+	bool IsBush(int soundType)
+	{
+		if (soundType >= 7 || soundType == 1)
+			return true;
+		else
+			return false;
+	}
+	
+	bool BushCheckCallback(IEntity entity)
+	{	
+		TreeClass treeClass = TreeClass.Cast(entity.GetPrefabData());
+		if (!treeClass)
+			return true;
+		
+		if (!IsBush(treeClass.SoundType))
+			return true;
+		
+		//2.5 because query entity sphere fucking lies, thanks BI
+		//Good balance between outside of bush/inside
+		if (vector.Distance(m_vOriginThisFrame, entity.GetOrigin()) >= 2.5)
+			return true;
+		
+		ApplyBushEffects();
+			
+		return true;
+	}
+	
+	void ResetBushEffects()
+	{
+		m_DamageManager.SetMovementDamage(0);
+	}
+	
+	void ApplyBushEffects()
+	{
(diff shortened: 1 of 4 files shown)


scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
```diff
@@ -0,0 +1,113 @@
...
+		//redundency to ensure this is NEVER on the server
+		if (Replication.IsServer())
+			return;
+		
+		ClearEventMask(GetOwner(), EntityEvent.FRAME);
+	}
+	
+	float m_fBuffer = 0;
+	override void EOnFrame(IEntity owner, float timeSlice)
+	{
```
tanaaKa, line 44: Super?

scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
```diff
@@ -0,0 +1,113 @@
...
+		m_vOriginThisFrame = owner.GetOrigin();
+		m_vOriginThisFrame[1] = m_vOriginThisFrame[1] + 1;
+		GetGame().GetWorld().QueryEntitiesBySphere(m_vOriginThisFrame, 1, BushCheckCallback, null);
+		if (!m_bEffectsAppliedThisFrame && m_bEffectsApplied)
+			ResetBushEffects();
+	}
+	
+	bool IsBush(int soundType)
+	{
+		if (soundType >= 7 || soundType == 1)
```
tanaaKa: All this needs to be is `return (soundType >= 7 || soundType == 1);`

scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
```diff
@@ -0,0 +1,113 @@
+class CRF_BushMovementComponentClass: ScriptComponentClass
+{
+}
+
+class CRF_BushMovementComponent: ScriptComponent
+{
+	[Attribute(params: "et")]
+	ref array<ResourceName> m_aBushPrefabs;
+	
+	float m_fOriginalSpeed = 0;
```
tanaaKa: I dont think this var is ever used.

scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
```diff
@@ -0,0 +1,113 @@
...
+			return true;
+		
+		ApplyBushEffects();
+			
+		return true;
+	}
+	
+	void ResetBushEffects()
+	{
+		m_DamageManager.SetMovementDamage(0);
```
tanaaKa: Missing `m_bEffectsApplied = false` here

scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
```diff
@@ -0,0 +1,113 @@
...
+		{
+			m_fBuffer += timeSlice;
+			return;
+		}
+		
+		m_fBuffer = 0;
+		
+		m_bEffectsAppliedThisFrame = false;
+		m_vOriginThisFrame = owner.GetOrigin();
+		m_vOriginThisFrame[1] = m_vOriginThisFrame[1] + 1;
```
tanaaKa, line 58: What about when players are uneven terrain? You sure this is accurate?

scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
```diff
@@ -0,0 +1,113 @@
...
+	}
+	
+	void ApplyBushEffects()
+	{
+		m_bEffectsAppliedThisFrame = true;
+		m_bEffectsApplied = true;
+		if (m_CharacterController.GetStance() == ECharacterStance.PRONE)
+		{
+			m_CharacterController.SetStanceChange(2);
+			SCR_HintManagerComponent hintManager = SCR_HintManagerComponent.GetInstance();
```
tanaaKa: You're getting the hint manager and hinting every frame lmao. Get the hint manager on init and send the hint on exit rather than every frame.

scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
```diff
@@ -0,0 +1,113 @@
...
+			m_fBuffer += timeSlice;
+			return;
+		}
+		
+		m_fBuffer = 0;
+		
+		m_bEffectsAppliedThisFrame = false;
+		m_vOriginThisFrame = owner.GetOrigin();
+		m_vOriginThisFrame[1] = m_vOriginThisFrame[1] + 1;
+		GetGame().GetWorld().QueryEntitiesBySphere(m_vOriginThisFrame, 1, BushCheckCallback, null);
```
tanaaKa, line 59: Big fucking yikes here imo. 10 calls a second lol, even more for higher fps players. We should exit early if the player is in a vehicle or not moving. IE:
```
if (m_CharacterController.GetMovementSpeed() < 0.1) return;
```

Going higher on the query would be good too

tanaaKa (changes requested):

scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
Salamiboio: Don't need it, it doesn't inherit anything beyond the script component

scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
```diff
@@ -0,0 +1,113 @@
...
+		TreeClass treeClass = TreeClass.Cast(entity.GetPrefabData());
+		if (!treeClass)
+			return true;
+		
+		if (!IsBush(treeClass.SoundType))
+			return true;
+		
+		//2.5 because query entity sphere fucking lies, thanks BI
+		//Good balance between outside of bush/inside
+		if (vector.Distance(m_vOriginThisFrame, entity.GetOrigin()) >= 2.5)
```
tanaaKa: Also aren't you checking this vector calculation on EVERY entity and even non bushes? I would imagine most entities would be non bushes.

Salamiboio: No we only check if it's a bush

`
if (!IsBush(treeClass.SoundType))
    return true;
`

is right above it to avoid unnecessary checks

tanaaKa: Did some more digging in engine. `Distance()` is most likely the basic sqrt formula of `sqrt((x2-x1)² + (y2-y1)² + (z2-z1)²)`.

We could probably save some calculations by doing:
```
const float BUSH_DISTANCE_SQ = 6.25; // 2.5 * 2.5
if (vector.DistanceSq(m_vOriginThisFrame, entity.GetOrigin()) >= BUSH_DISTANCE_SQ)
    return true;
```

Salamiboio: Probably, I do this for CVON, it literally did nothing to frame times but I can add it

tanaaKa: Yeah I mean we're really getting nit picky about it tbh, but aim for the moon and all that.

scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
Salamiboio: You can prone while moving so we are almost always forced to be checking

Salamiboio: You can prone while not moving*

tanaaKa: TRUEEEEEEEEEEEEEEEEEEEEEEE

Salamiboio: Also we use delta time here so it's always gonna be 10 times a second fps aside

scripts/Game/Systems/Components/Character/CRF_BushMovementComponent.c
Salamiboio: Yeah positive, origin is always at the feet and the character is a fucking mountain goat

Evidence of said mountain goating
image

Salamiboio:
Also fixed a bug with taking damage in the bush, should be gtg now

PR #1121, opened 2026-01-02
Stats Enablement for 2026

tanaaKa:
Enables the datacollector for our game mode so we collect the same stats as conflict while keeping our original stats system


Salamiboio:
Does this already work with our old discord bot like it used to?

tanaaKa:
This enables the .playersave feature that reforger uses in the %PROFILE% directory. It doesnt touch what we're doing externally with logging.

Njpatman (approved):
LGTM

PR #1124, opened 2026-01-03
RPL Maintenance and Entity Improvements

tanaaKa:
Fixes issues that have been seen in client and server error.logs, compiler warnings about obsolete methods and incorrect upcasting, and the initial entity not deleting when a player assumes a playable entity


scripts/Game/Systems/Core/Containers & Configs/CRF_SightArsenalRegistry.c
Njpatman, line 1: Please move the registry to either a new "scripts/Game/Systems/Core/Registries" folder or to the "scripts/Game/Systems/Core/Containers & Configs" folder

scripts/Game/Systems/Core/Managers/CRF_GamemodeManager.c
```diff
@@ -457,10 +469,44 @@ class CRF_GamemodeManager : SCR_BaseGameModeComponent
...
+			// Log deletion for debugging
+			Print(string.Format("[CRF] Deleting ghost initial entity for player %1 at position %2", 
+				playerController.GetPlayerId(), 
+				oldEntity.GetOrigin()), 
+				LogLevel.VERBOSE);
+			
+			// Delete immediately to prevent replication
+			SCR_EntityHelper.DeleteEntityAndChildren(oldEntity);
+		}
+	}
```
Njpatman, line 508: I don't believe this is needed.

In CRF_PlayableCharacter.c, we have a method named "ShouldDeleteEntity" that is checked every frame to determine whether we should delete the initial entity or not. We should double-check and ensure that it isn't firing before we add a whole separate method. And if it is firing and we've verified the entities still exist, I'd suggest looking into that rather than going through the gamemode manager

Njpatman (changes requested):

scripts/Game/Systems/Core/Managers/CRF_GamemodeManager.c
tanaaKa: This method runs complementary to the already implemented `ShouldDeleteEntity()`.

IE, `DeleteOldInitialEntity()` runs linear and in our expected case where it always cleans up the last entity after possessing happens, but `ShouldDeleteEntity()` runs on each frame and will clean up any edge cases. 

I would say our initial implementation of `ShouldDeleteEntity()` is blanket-based, but the one I wrote above forces the spawn/respawn of entities to happen in a specific way no matter what rather than coming in afterwards with a blanket fix.

Njpatman (approved):
LGTM

PR #1140, opened 2026-01-12
Fight Club Add. + Changes to MapStaging.c, PolyZoneTrigger.c, Looter.c

HarryEngineering:
Making some improvements to the special.

- [x] Rebalance medical distribution
- [x] Add a default tourniquet
- [x] Cleanup missing spawns
- [x] Adjust zone visibility
- [x] Add a very small final zone, and redistribute timing
- [x] Fixed Parachutes not deploying
- [x] Added individual unit markers for the player and his group members


TristGit:
- [x] Changed it to 5 zones, made visible, and adjusted timers (23~ minutes total)
- [x] Added some objects+spawns to final zone
- [x] Changed Scripts:

**PolyZoneTrigger.c** - Added an optional independent 'Search + Apply Boundary Effect To Outside Players' method nested within the class, for reversed trigger zones. Did this because the polyzone triggers even when reversed do NOT detect/apply OOB players until they update by entering. Must to be called/casted to. Isnt part of query loop + only when using reversed zones as well. Safe. This fixes an actual issue with my mapstagingcomponent but can be called/casted to by anything.

**MapStagingComponent.c** - Added "Delete Previous options, and added above CRF_PolyzoneTrigger functionality

**LooterGamemodeComponent.c** - BI changed display names for containers, fixed to label spawn config item names correctly.
 
 Tested on dedicated, all works fine, only oversight is typos/oversight on settings I forgot to change back when testing I think

Salamiboio:
Keeping the deployable parachutes, separate pull request is going to cover my fix for the parachutes not deploying.

Salamiboio:
Added markers for players in the same group so they can see where each other are at, parachutes are fixed in https://github.com/CoalitionArma/Coalition-Reforger-Framework/pull/1151 and still remain as a manual deploy for this

PR #1153, opened 2026-01-18
De-Factionize Gearscript Characters

Njpatman:
We now use the faction affiliation comp for each character (this is more like how vanilla handles character spawning).

This also significantly reduces the size of configs we search through for players' resources, which should lend a small boost to the gearscript manager's performance.


Njpatman:
@tanaaKa ready for review, HEAVILY recommend using the file filter to just filter out the .meta/.et files (those are a byproduct of having to regenerate the meta files for the groups, ig it decided to update literally every meta file lmao)

scripts/Game/Systems/Core/Managers/CRF_GearscriptManager.c
```diff
@@ -265,25 +260,14 @@ class CRF_GearscriptManager : ScriptComponent
...
-				return "OPFOR";
-			case resourceName.Contains("INDFOR"):
-				return "INDFOR";
-			case resourceName.Contains("CIV"):
-				return "CIV";
-		};
-			
-		return "";
+		FactionAffiliationComponent facComp = FactionAffiliationComponent.Cast(entity.FindComponent(FactionAffiliationComponent));
+		return facComp.GetAffiliatedFactionKey();;
```
tanaaKa: Double semicolon

scripts/Game/Systems/Core/Managers/CRF_GearscriptManager.c
```diff
@@ -265,25 +260,14 @@ class CRF_GearscriptManager : ScriptComponent
...
 	//------------------------------------------------------------------------------------------------
 	/**
-	 * @brief Determine faction key from resource name
-	 * @param resourceName Resource name to analyze
+	 * @brief Determine faction key from faction affiliation comp
+	 * @param entity Entity to pull the faction comp of
 	 * @return Faction key or empty string if not found
 	 */
-	protected FactionKey DetermineFactionKey(ResourceName resourceName)
+	protected FactionKey DetermineFactionKey(IEntity entity)
```
tanaaKa, line 267: I know the faction manager *should* never be null, but you really should have a null safety check here on facComp

tanaaKa (changes requested):

PR #1160, opened 2026-01-22
Updated Fight Club / Add BattleRoyale Component

HarryEngineering:
- Moved everything to the new slotting method, rebalanced shotgun drop rate.


TristGit:
- Added BR component. It handles: Win Screen, Alive Player count, Zone RNG
(I disabled MapStaging for this, there is a chance the replication involved is not optimized well enough but I tried)
XRTjPn

- 4 New possible Zones
- New props for some of the zones
- Removed Signal Hill zone (cause it was played 4x)
- Disabled Nametag component
- 5 minute zone timers, 150 second initial for prep after safestart for zoning

3zWEFt



- [x] Any cursory testing with short 20s timers set in the zones (can also force a zone with the debug setting) appreciated.
- [x] Glance over settings, anything I may have left on in the CRF_Lobby while testing

Tried my best to optimized both client and replication side of the component, basically constantly tracks time for some player count / winner detection functions.


Essentially an almost done protoype.

PR #1168, opened 2026-01-26
Adds Camera Manager

Njpatman:
Also fixes cameras not going into AO center


Njpatman:
No actual changes to the code were made, just where we parse it.

PR #1171, opened 2026-01-27
Bug Fix | Fixes Parachutes For Real

Salamiboio:
Fixes players not being able to deploy parachute near the ground.
Removes the ability to break your legs/die if you hit the ground too hard.
Updates it for the latest version of the parachutes.
Re enabled manual deploying for fight club


Salamiboio:
Gonna approve this so we can work in the CRF again

PR #1184, opened 2026-02-10
BR update

HarryEngineering:
Update USSR 80s gear config: swapped Backpack_Kolobok_Parachute_Chicken.et for the standard Backpack_Kolobok_Parachute.et (GUID updated), and removed the CRF_Inventory_Item entry for ACE_Captives_ZipCuffs (count 2). This removes the chicken parachute variant and the zip cuff accessory from the unarmed loadout.


TristGit:
Will append some of my own changes into this later today.

PR #1191, opened 2026-02-15
Crf_HVT Mode Rework

TristGit:
Designed to allow for multipled hvts. Nobody will be surprised or phased as it technically what we already do just in one script
- Allows for multiple player VIPS/HVTs, different sides, objects, AI (previous features)
- Previous component updated the origin location w/BumpMes every second to client/server; Now RPL frequency is locked behind whatever the Ping timer is set to + a repeating 60s server side re-registration of HVTs (JIPs, DCs, or Arma'd respawns)
- also have 2 external methods that are meant to assist with counting or finding nearby hvts

- Updates the only dependent mission (Cass's SiliconValley)
- Removes scripts that Harry said to just remove as they were dependent on previous single-variable prefab format

image


Njpatman (approved):
LGTM

PR #1193, opened 2026-02-16
GetSlotData: O(n) to O(1)

tanaaKa:
- Greatly reduces the complexity of the slot data updates
- Instant visual feedback when taking slots


tanaaKa:
Moving to draft so I can expand this

PR #1197, opened 2026-02-19
Feature | Separates factions in spectator

Salamiboio:
If Respawns are enabled, you will no longer be able to hear the enemy faction in spectator.


tanaaKa:
Are you positive this works? I was under the impression that dying removes your slot from the slot list, thereby resetting your status.

tanaaKa:
Also we dont want this for normal missions. Needs a way to separate based on game mode component.

Salamiboio:
Positive, tested and works, dying keeps all your slot data still (Its how you respawn into the same slot)

If you wanna test it yourself. Open up vondata.json in VSCode and watch the attenuation values.

Njpatman (changes requested):
> Also we dont want this for normal missions. Needs a way to separate based on game mode component.

^This, no need to hook it into the plugins, can just be a variable on the gamemode.
Please ensure this is off by default, since this will primarily only be used in events/long form missions

Njpatman (approved):
LGTM

PR #1200, opened 2026-02-21
Keep the slot list the same when respawn is enabled

tanaaKa:
Title


Njpatman (changes requested):
Would highly recommend simply disabling the hiding of dead slots on the actual slotting screen when respawn is enabled. Many systems rely on getting a person's "dead" status from their slot, so this may break things unintentionally.

tanaaKa:
Ok so how

Njpatman:
> Ok so how

Scripts\Game\Systems\UI\Menus\CRF_SlottingMenu.c
Line 761 - Method: "AddSlotsToGroup"

Inside that method on line 787, you can add a simple check to see if respawn is enabled to show dead slots, or you can remove that check entirely.

Salamiboio:
Id say to completely remove it, I didnt do this in first iterations of the lobby because it can be used to gather meta info if you see a bunch of slots gone.

tanaaKa:
Added

Njpatman (changes requested):
Image

This is what I see? i don't see any changes to the slot list

tanaaKa:
You dont see m_SlottingManager.UpdateDeathState being removed?

Salamiboio:
> You dont see m_SlottingManager.UpdateDeathState being removed?

Don't remove that. Death state in the slot is how we track the death of an entity without having to find it in the world. 

Just remove the code that hides it in the client UI

PR #1204, opened 2026-02-23
Add Ability To Set Friendly Factions

Njpatman:


scripts/WorkbenchGame/MissionPlugins/CRF_MissionFactionPlugin.c
```diff
@@ -9,32 +9,47 @@
 ] 
 class CRF_MissionFactionsPlugin : WorkbenchPlugin
 {		
+	// Due to limitations in getting the faction managers faction list (obfuscated) we have to manually point to each factions index.
+	static int BLUFOR_FACMANAGER_INDEX = 12;
```
tanaaKa: Are these indexes going to change when we remove/add faction mods? If so, we need a better solution.

Njpatman: Yes, CIV will never change but all other factions will, the problem is there is no way of actually getting the raw faction array since it's obfuscated

Njpatman: Hmmm, may be able to do something, since all our custom factions are added at the bottom of the array in the same order each time due to them being added in the CRF_LOBBY Instance.

Njpatman: Fixed in the latest Commit

Njpatman:
Ready for re-review

PR #1205, opened 2026-02-23
Parachute System Implementation

HarryEngineering:
This PR adds a simplified parachute system to the game, consisting of three main components:

    CRF_ParachutePlayerComponent – Attached to player controllers. Handles deployment checks (altitude, obstacles, falling state), spawns the parachute entity, transfers ownership, and manages safe deletion after landing.

    CRF_ParachuteBackpackComponent – A simple component on backpack items. Tracks whether the parachute has been used and stores the prefab to spawn.

    CRF_ParachuteDeployedEntity – A custom entity that represents the deployed parachute. It simulates descent (with drag limiting fall speed), applies wind drift, detects ground contact, and requests exit when landed.

Key Features

    Deployment Conditions – Player must be falling (PhysicsIsFalling()), above minimum altitude, and clear of obstacles.

    Safe Deletion – After landing, the system waits for the cargo slot to be empty before deleting the parachute, preventing player deletion while inside.

    Wind Simulation – The parachute receives a continuous impulse based on global wind direction and speed, affecting its trajectory.

    Network Synchronization – The server broadcasts the parachute's transform and velocity; clients interpolate for smooth movement.

    Retry Limits – All retry loops (entity resolution, entering, ownership) are capped at 10 attempts to prevent infinite loops.

    Deletion Protection – Network RPCs and interpolation are ignored once the parachute has landed or its physics component becomes invalid, preventing errors when the entity is deleted.


tanaaKa (dismissed):

HarryEngineering:
There's an edge case where you can either get dropped too high or low based on terrain and it'll break your legs or drop you under the terrain. Working on that.

HarryEngineering:
This makes fight club ready to play from a parachute standpoint, so I'm going to go ahead and make it ready for review. Model, BR changes, etc can be in a second pr. We're likely to play the new rally first tomorrow, but if we could give this a test after I'd appreciate it. I'm sure there will be things to fix and tweak.

PR #1209, opened 2026-02-24
Feature | Adds a radius to pointing

### LINKED ISSUE #1202: [CRF Feature] Chanel ACE Point icon from a direct icon to a circle
Should be as easy as finding what layout they use for the marker and override it

Salamiboio:
Fixes https://github.com/CoalitionArma/Coalition-Reforger-Framework/issues/1202

Adds a radius to where you point now so there's more room for interpretation then just a straight shoot here. Muh realism or something.

image


Salamiboio:
Also lowers the time it takes for a point to disappear from 6 to 3 seconds

Njpatman (changes requested):
Please move any prefabs out of prefabs editable and move them to either a new or one of the folders under "Prefabs"

Otherwise, LGTM

Salamiboio:
Fixed the above issue

PR #1215, opened 2026-02-25
Feature/spec maximizing

### LINKED ISSUE #1128: [CRF Feature] Spectator Follow Camera Expansion
The old reforger lobby system's spectator included the ability to fix spectator camera to a player. This would make your camera follow players _without_ locking your view eg. the current helmet camera. You could still reposition the camera while keeping it moving relative to the player. The current spectator system, while functional, lacks this and could be expanded in multiple ways.

This functionality should be returned, and additional functionality should be investigated such as the ability to move the spectator camera in relation to other entities (eg. vehicles, drones, helicopters). This would massively expand the types of dynamic shots spectators would be able to capture.

tanaaKa:
Adds a multitude of features to spectator:

- 3pp orbiting "latch" camera
- player information panel when directly spectating
  - player status information tied to ace medical
- dead player names now appear on dead bodies
- group nato symbols organized by side/color
- fading on both group and character icons
- probably some others im forgetting

image
image
image


Salamiboio:
Fixes https://github.com/CoalitionArma/Coalition-Reforger-Framework/issues/1128

Njpatman (commented):
All I ask is that the followzoneHud be made into a modular layout element with a attached script so you can just spawn the layout when needed instead of manually creating and destroying widgets as you select/deselect a player.

I'll get it done tonight myself if that's acceptable.

tanaaKa:
image

Added uncon and resiliency percentage for wakeup

PR #1224, opened 2026-03-01
FightClub SPCL - Failtest For Session (Temporary)

TristGit:
Im putting this out before Sun. session just as a potential to run before/after to 50/50 the failpoints of the FightClub.
Just disabled the BRComponent and placed down a marker for convergence.
Not permanent.


TristGit:
Failtest w/out the BR component and zones resulted in same crash happening within 5-10~ minutes after drop

PR #1238, opened 2026-03-04
Bug Fix | Fixes vehicles not updating when moving the spawner

Salamiboio:
- Adds comments (Fuck njpatman)
- Also fixes a bug when you moved the spawner in workbench the vehicle pos didn't update


Njpatman (approved):
All the comments belong to us.

PR #1242, opened 2026-03-04
New Mission- Harry DOM Lionshead

HarryEngineering:
Mission Pull Request Template

This Pull Request will:

Add the mission Lionshead made by Harry to the repository.

- [x] Mission File
- [x] Gearscripts Tested
- [x] Side-based briefings  
- [x] Mission .conf file

**Faction(s):**

BLUFOR - Modern US Defending
BLUFOR Comp: Inf
BLUFOR Assets: Entrenchment, 1x AP Mine Laying MRLS

INDFOR - Modern ION Attacking
INDFOR Comp: Inf
INDFOR Assets: Sniper team, Drone team, 1x Smoke MRLS

Attack or defend the objectives in order. Blufor can get supply to build with from the forward point.

**Terrain:**

Zimnitrita

ION has been contracted to push the new American force out of Varpadany before they can get fully built up. 

**Slotting Ratio:**

4:5

**Time limit:**

<60min>

image

There's two capture triggers, and new gearscripts.

**Design concept/thoughts:**

Wanted to make a modern hard fight without vics.

**Other Notes:**

Indfor really needs the specialties, I think.


HarryEngineering:
I'm going to add a grad to each side, and trim the AO on the defending side.

HarryEngineering:
Gave each side a utility grad, with a teleport pole to it. Changed the AO border to minefield markers to allow teleport out.

shimavitz:
MLRS is out of safe zones, is basically an arty asset without a dedicated slot, is the intent that commanders go do an artillery firemission?
The northern most objective is outside of safestart, is the intent to delay blufor building activity?

HarryEngineering:
> MLRS is out of safe zones, is basically an arty asset without a dedicated slot, is the intent that commanders go do an artillery firemission?

You can get to the MRLS by teleporting via the poles. It's explained in the side briefing. Yes, it doesn't have a dedicated slot because they are one time support options without direct kinetic killing potential. I'll leave it up to PL to manage how to man it. 

> The northern most objective is outside of safestart, is the intent to delay blufor building activity?

Yep. Blufor has some starting supply, and they can build as many trenches as they want, but they should be incentivized to hold the forward point as long as possible.

shimavitz:
Id rather you use the RHS CISS instead

shimavitz:
IDK if I like defenders not being able to go to their first defense point

shimavitz:
*during safestart

HarryEngineering:
> Id rather you use the RHS CISS instead

No thanks, that system has it's own issues to resolve. If it's a problem I'll just revert the addition. It was only added because MM were asked to use new TF stuff.

> IDK if I like defenders not being able to go to their first defense point

Is there a balance or immersion breaking reason why? I don't have a strong opinion on this one.

HarryEngineering:
Thinking about it, I'll add indirect leads to each platoon element. That way at least someone can know what they're signing up for, and if they don't get slotted then pl can figure it out.

shimavitz:
QA
indfor medics and arty man have only 1 mag no spares
give blu spear mcx users at least 2 more mags (TLs AARs)
communicate exact muniton type for the mlrs (large brake ring vs small break ring)

shimavitz:
the mlrs are allowed to move correct?

shimavitz:
VIP mini arsenal should probably not be a thing.

Changing to some weapons on AAR in arsenal removes the mags for your AR's weapon AKS-74U for example but i think i made it happen with others

maybe give one starting utility truck in their spawn, even though there are some alternative transports elsewhere, maybe some more spare cars around the AO because it is a fairly large play area

Breacher shotgun only has the one "mag", not sure if intentional
Nods do not always seem to work properly, not sure as to why but pressing N wont work.

HarryEngineering:
> the mlrs are allowed to move correct?

Yes

applied QA feedback

PR #1264, opened 2026-03-07
Standalone UI modules

tanaaKa:


Njpatman (changes requested):
Two things:

1 - The "Ding" when switching slotting phases no longer exists

2 - The actual UI layouts need to become modularized themselves. Otherwise this doesn't really change anything, just separates some scripts out. Look at how I did the EntityInfoDisplay.layout as an example.

Njpatman:
This is dead, unless you want to revive it in lobby @tanaaKa?

PR #1266, opened 2026-03-07
Slotting map updates

tanaaKa:
Batches individual slot updates
Fixes UI flashing

### DIFF
--- a/Scripts/Game/Systems/Core/Managers/Gamemode/CRF_SlottingManager.c
+++ b/Scripts/Game/Systems/Core/Managers/Gamemode/CRF_SlottingManager.c
@@ -221,10 +221,17 @@ class CRF_SlottingManager : ScriptComponent
 			CRF_FactionSlotStats stats = m_mFactionStats.Get(factionKey);
 			if (stats)
 			{
+				// Update locked count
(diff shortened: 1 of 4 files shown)


Njpatman (changes requested):
Image

- The total number of players slotted/available slots under each faction flag doesn't update in the slotting screen.

Image

- Players can change the slotting ratio calc on the fly, and it just bricks the slotting screen if they do lol

Image

- Dead slots aren't showing again on the slotting menu (Company Leader)

tanaaKa:
Should be all updated now

Njpatman (changes requested):
Image

 - The Slotting menus are noticeably slower, with locking a squad now going up to 6000us peak in just the slotting menu class, with the broadcast manager right behind.

In reference, Release has a peak of around 1300us when locking a squad:
Image

Image

 - Slotted/available under each faction flag is still semi-broken, can't see the actual slotted players part of the [Slotted Players/Available Slots].

 - Also, slotting ratio changes don't update on all clients, just on the local client that made the change

tanaaKa:
Just going to abandon since it's slower

PR #1271, opened 2026-03-12
New Mission | tanaKa HVT80 NO MOTION

tanaaKa:
- [x] Mission File
- [x] Gearscripts Tested
- [x] Side-based briefings  
- [x] Mission .conf file

**Faction(s):**

BLUFOR - Zimnitrita SF - Attacking
BLUFOR Comp: Special Forces + Heli Support
BLUFOR Assets: SF Helis

OPFOR - Euro Insurgents - Defending
OPFOR Comp: Infantry
OPFOR Assets: Random civilian vehicles

CIV - (EX. Civvies) - HVI
Captured by OPFOR at start

Both sides need to eliminate the other and/or extract the HVI in the process.

There are numerous possible outcomes to this mission:

HVI killed - no sides extract - both sides fail
HVI killed - both sides extract - draw
HVI killed - one side extracts - extraction victory
HVI traded - draw
HVI extracted - extraction victory

The only true victory is one where the HVI is extracted and either both sides extract or one side is eliminated prior to extraction.

**Terrain:**

Zimnitrita

OPFOR has captured an informant for BLUFOR and BLUFOR has opted to attempt an extraction rather than abandoning him.

**Slotting Ratio:**

BLUFOR 2:3 OPFOR

**Time limit:**

55m

image

None. Uses the HVT game mode system.

**Links to Added Mods**

[Stun Grenades](https://reforger.armaplatform.com/workshop/59EAA899751805DF-StunGrenade)
[H-47 Chinook](https://reforger.armaplatform.com/workshop/61957C5C6FB7A773-H-47Chinook)
[TroopFit Arsenal (Vergys New Clothes)](https://reforger.armaplatform.com/workshop/686146A940C9D3EA-TroopFitArsenal)

**Design concept/thoughts:**

This is a non-standard asymmetric mission where the attackers are better equipped than the defenders, but the defenders have terrain and number advantage. Previous missions of this type have been well received. BLUFOR has many magnified optics that are balanced by disabled their nods while aiming. BLUFOR also only has stun grenades rather than frags. Leadership also lacks M203s to prevent fragmentation.

BLU is given numerous types of aircraft to choose how they wish to complete their mission.

OPF is given three separate play areas they can choose to forward deploy to in order to keep the mission replayable. OPF is given two PKM uazs at the beginning that can be given as needed and each deployable area has plenty of civilian vehicles they can utilize.

BLU will see the area chosen via the HVIs transponder that appears on BLUs maps. BLUs main force multiplier are the miniguns on rotary wing as well as the AH-6M camera for recon.

**Other Notes:**

N/A


shimavitz:
VIP mini arsenal should probably not be a thing.

Changing to some weapons on AAR in arsenal removes the mags for your AR's weapon AKS-74U for example but i think i made it happen with others

maybe give one starting utility truck in their spawn, even though there are some alternative transports elsewhere, maybe some more spare cars around the AO because it is a fairly large play area

Breacher shotgun only has the one "mag", not sure if intentional
Nods do not always seem to work properly, not sure as to why but pressing N wont work.

Salamiboio:
Make sure the nods are placed in default inventory and not attached to helmets in prefabs. Still a bug in RHS.

tanaaKa:
> Make sure the nods are placed in default inventory and not attached to helmets in prefabs. Still a bug in RHS.

Shutup nerd
image

tanaaKa:
> VIP mini arsenal should probably not be a thing.

Im okay with leaving it on so the VIP can customize his look each time the mission is played.

> Changing to some weapons on AAR in arsenal removes the mags for your AR's weapon AKS-74U for example but i think i made it happen with others

Gearscript bug

> maybe give one starting utility truck in their spawn, even though there are some alternative transports elsewhere, maybe some more spare cars around the AO because it is a fairly large play area

Im okay with giving them a starting truck, but no one should be driving out of the spawn area. All squads should be deploying in order to maximize their fortification time. A truck would be counterproductive to this and there are trucks at each area already.

> Breacher shotgun only has the one "mag", not sure if intentional

Yeah that's fine. It's a limitation on the gearscript system since breachers are technically HAT gunners and the shotgun is their hat weapon. Shotguns are op so probably for the better in this case.

> Nods do not always seem to work properly, not sure as to why but pressing N wont work.

Also a gearscript issue. Sometimes happens after changing gear in miniarsenal as well. If nods are in inventory, dropping and reequipping them is the workaround.

tanaaKa:
I think this mission hinges on smart recon by the BLU PL before action starts. Maybe a strobe in the VIPs inventory would better help that. Adding now.

tanaaKa:
Added CIV IR strobe, removed hardhats and rain coats from uniform options, set to standalone gearscript rather than base CIV script.

PR #1275, opened 2026-03-12
TrySpawnPrefabToStorage

tanaaKa:
Test PR for TrySpawnPrefabToStorage


tanaaKa:
@Salamiboio @Njpatman Thoughts?

tanaaKa:
On second thought, will push into release for thorough testing during todays coop. Already tested functioning on server.

Njpatman:
You need to spawn weapons in as weapon proxy's, will cause the weapon disappearing out of rpl bubble range to happen again.

Njpatman:
There should be a way to get the type of proxy to spawn on the actual entity, may need to investigate the vanilla arsenal methods

Salamiboio:
It also deletes the attachments block. Recommend keeping weapons the exact same and only using this for regular inventory items.

PR #1290, opened 2026-03-18
Inventory Spawn Changes

tanaaKa:
As requested


tanaaKa:
Verified functioning in todays coop

PR #1292, opened 2026-03-19
New Mission Quack TVT M777 Heist

mnavarrete1017:
Mission Pull Request Template
PR NAME:
New Mission | Quack TVT 129 M777 Heist

This Pull Request will:

Add the mission M777 Heist made by Quack to the repository.

- [x] Mission File
- [x] Gearscripts Tested
- [x] Side-based briefings  
- [x] Mission .conf file

**Faction(s):**

BLUFOR - US 2000's UCP - Attacking
BLUFOR Comp: Motorized Infantry
BLUFOR Assets:Trucks, M2 HMMWVs, MMG Team

INDFOR - MEI - Defending
INDFOR Comp: Infantry
INDFOR Assets: 2x M777, Recon Team, AT Mines, Special equipment

BLUFOR wins when it moves both M777s to their drop off point. 
INDFOR wins if it moves Both M777s to their drop off point.
Tie if both faction has 1 M777 on their drop off point when time ends.

BLUFOR: Assault Murabat and retrieve the M777s
INDFOR: Defend Murabat for 20 minutes and then move the M777s using the trucks provided.

**Terrain:**

Anizay

An American Artillery battery has been ambushed by INDFOR and lost two of its guns. US forces head to the town of Murabat to retrieve them.

**Slotting Ratio:**

5:4 Favoring BLUFOR

**Time limit:**

45 Min
image

• There are two civilian trucks with a custom component(Quack_TimedVehicleTeleComponent) that should teleport them to the INDFOR spawn after 20 minutes.
• There's a wave respawn for both factions set to 20 minutes as well.
• The BMP in the unnamed hamlet has been converted to a storage container. Test if player clients can "open" the storage located near the rear doors.

**Links to Added Mods**

None

**Design concept/thoughts:**

I think the town of Murabat is a great place to defend. Players will not be provided with a range table and are instead encouraged to use the M777 as a direct fire weapon.

**Other Notes:**
INDFOR may find sandbag parts in the back of the captured US trucks.
INDFOR may find extra RPG75 in the unnamed hamlet west of Murabat. 
INDFOR may find NSV parts and two boxes hidden underneath the bushes within the Hill House compound. 

M777s are finicky when pushed. They have a tendency to run away from the user.

If the towing truck and the M777s are not oriented in the same direction when the tow action is used, there's a chance that the M777 will explode.


HarryEngineering:
I really like the commander's notes. Very cool mission idea fitting in theme with the sandbox. Looking forward to it. 

This is a checklist for you to fill out showing you've checked those things ahead of submitting the PR. 

image

What's to stop the Indfor team from interdicting blu on the way, and leveling them with the m777? Maybe change out the charge 1 for charge 4. It'll be much more difficult to stay in the engagement range for indirect, and it doesn't matter for direct fire. 

This seems like a very hard fight for blufor. Near even ratio when they have to do the maneuvering through hostile territory with lots of concealment/cover, and the enemy has a lot of comma's in their special assets that blu has to overcome with GI, humvees, and an mmg team. Hard fights are fine if everyone is engaged, such as in Tanaka's hard trench assault a few months ago. 

Similarly, transport trucks are generally an indication in a short 1 life that there's something not quite right. There's a lot of terrain to cover, and trying to do it in a truck introduces a no-win situation for blufor. Either they take the truck all the way in, and dismount in hostile controlled territory, or they dismount before like they are supposed to, and then run short on time walking. You have to be careful with the solution too. If you increase the timer to allow for foot travel, that's a lot of empty maneuvering time. Try to come up with something to address this. 

Recon doesn't need a team leader, the Sniper is the leadership for that element. 

Need to fill out time (mission duration) and weather. For weather I like to include TOD, so Midday Clear. 

image

HarryEngineering:
You also want to indicate where the replacement trucks spawn. I thought based on the briefing language they would spawn in town.

mnavarrete1017:
The M777s are located in a depression. If INDFOR wants to do indirect fires they would need to use plunging fires, otherwise the rocky terrain would get in the way. This will be very difficult without a ranging table guiding the initial shot. Coupled with only 6 shells in the trucks it would generally be a waste of munitions to attempt. I've decided to replace the 1 charge to the 5 charge.

The trucks are more than people movers in this mission, as they will be required to tow the M777s back to BLUFOR start. Whether they move on foot or try to tail blaze their way to the gates of Murabat, those trucks will need to come with them. For that reason I've decided to give BLUFOR some extra just in case they get into an engagement that results in their destruction. I could spawn more if you think 4 is not enough, or give them an weaponless APC; perhaps an M113?

I hesitate to give BLUFOR anything heavier than HMMWVs, like a LAV. This may cause an awkward arms race where I'll need to give INDFOR RPGs, and considering the terrain, there are some locations where the entire convoy could be destroyed. Should I give BLUFOR more M2 HMMWVs? As I told creme monday, It takes about 2 minutes and 30 seconds to arrive to a potentially INDFOR controlled area (about halfway); 17 and a half minutes is a lot of time to engage INDFOR. 

Currently INDFOR does not start with AT assets. They'll need to forward deploy to gain either RPG75s or an NSV machine gun; as a consequence this will leave the town less defended. INDFOR may attempt to ambush BLUFOR on route, but since there are two routes into Murabat, if the ambushers are bypassed, it's a long walk back to town. I've decided to bring the NSV box to 1. 

The replacement trucks will/should spawn right in front of INDFOR spawn in the empty lot next to the forested area.

HarryEngineering:
Ok. Make sure to emphasize blu will need two transports trucks to tow the m777 when Tanaka asks you to give a brief about the mission before slotting. I don't have specific recommendations about assets because I think it's a good thing to have variety in MM style. My comments about balance aren't show stoppers. 

Make sure the trigger works in the dedicated server tool, I only saw trucks at the extract when I waited for it earlier and assumed those were the spawn. If you don't know how to use the dedicated server tool, reach out in the discord. 

Don't forget to change the recon element, and then delete and recreate the config file for the new player count.

Njpatman (changes requested):
Please do not make changes to base gearscripts without a good reason; simply make a variant in your own separate personal folder.

PR #1306, opened 2026-03-23
TrySpawnPrefabToStorage Integrated Into Gearscripts

Njpatman:
Does what it says; this significantly reduces load time since we are no longer spawning the item/clothing piece on the ground, then inserting it into the player's character. We just spawn it directly into the player's inventory.


Njpatman:
Ready For Review

PR #1309, opened 2026-03-23
Get rid of JSON buffer

tanaaKa:
Do not merge until https://github.com/CoalitionArma/Coalition-VON/pull/31 is complete.

Changes JSON writing to be 60hz ~16ms compared to 50ms.


Salamiboio:
We buffer it cause the TS plugin is limited to reading every 50ms so as not to stall it.

tanaaKa:
Like an arbitrary limitation on the c code in the plugin/teamspeak? Or did you just choose 50ms by default? From what I can tell, the worker thread in `plugin.c:2056` sleeps for WORKER_TICK_MS (200ms) at the end of every loop iteration, in 50ms slices (those slices are only to allow a fast quit, not for polling). The VON reload gate if (now >= g_nextVonReloadTick) uses VONDATA_RELOAD_MS = 50, but since the loop already slept 200ms, that gate is always expired when it's evaluated. The effective VON check rate is 200ms, not 50ms.

IE if I set `VONDATA_RELOAD_MS` to 1ms or 200ms, it would produce the same behavior.

tanaaKa:
Furthermore, just checked TFAR. That implements 100ms on the worker thread, so we were 2x slower than tfar.

Salamiboio:
Hmm, yeah that seems to be a bug, god who wrote this shit.

Ill update the worker thread to be 100ms and also get rid of the wet sock effect with behind intensity lmao.

tanaaKa:
With my PR I have it down to the physics tick at 60hz/16ms. Give that a try first. Enfusion should be able to handle faster report.

As well as a bunch of other cleanup.

Salamiboio:
Check the frame times in your PR if you can, I used to do this every frame and it had some insane frame times.

Salamiboio:
Wrote the buffer because in theory we shouldnt need to process data faster than the TS plugin can handle it

tanaaKa:
My branch:
image

CVON main/CRF release:
image

Salamiboio:
Hmm, poggers

tanaaKa:
@Salamiboio Reminder. Needs done sooner rather than later.

PR #1313, opened 2026-03-25
Add a real rally component

tanaaKa:
Adds a rally component with an example world to the repo.
- Rally start and end system
- Lap support
- Checkpoint system
- Singles and duo (team) support
- Adds ending rally screen
- Adds mid-race screen with times and players
- Adds chat updates on player race status(es)

Tweaks the S105_rally car prefab to have better power, transmission, suspension, and tire grip.


Njpatman (approved):
The only real critique I have is that we should start using entities/components that automatically register with the gamemode, rather than having to manually name entities. It makes the mission-building process much smoother. Otherwise, LGTM

PR #1363, opened 2026-04-15
Adds chat command to open admin menu

Salamiboio:
/adminmenu in chat now opens the admin menu in case a bug causes the left side of your screen to no longer function.


tanaaKa:
May I request you also add a `/r` command to respond to the latest ticket?

Salamiboio:
We already have that, just gotta know the player id which is attached to the ticket. Probably could change to the name honestly

Salamiboio:
Switched the command over to player names now

PR #1417, opened 2026-05-05
New HVT Update v1

TristGit:
Fixed(?) Client Desync experienced last week, Maintained Perf. Checks, Added Harry's hvt transponder alert + Toggle Also changed visual of hint layout to match UI.

**Optimizations**
- `m_mEntryToTransponder` caching, `SetHVTAndState`, no repeated `FindEntityByName` in sync loop
- `m_mHVTDamageManagers` caching, `IsHVTAlive` does O1 lookups instead of repeated `FindComponent`
- Player registration check 60s changed to 120s
- Fixed transponder marker `timeDelay` set to `m_timeBetweenPings` causing marker desyncs

**Additions**
- Toggle for CRF Hint alert for all/searcher faction every time the HVT tracking updates
- `m_bHVTPermadeath`, locks a PHVT index entry behind `m_sPermadeadEntries` on death, blocking re-registration. They just become just another soldier. Prevents incidental respawns just resulting in HVT's showing back up/being pinged
- `/hvt_reset_deaths`, admin chat command clears `m_sPermadeadEntries` For oopsies or hiccups on HVT's parts, lets HVT's be re-registered.


tanaaKa:
I have some performance concerns with the HVT game mode in general so this might be a good PR to address or tell me no:

- `SyncTransponderPositions()` calls `GetGame().GetWorld().FindEntityByName(...)` for every HVT entry on every ping cycle, and on every client. FindEntityByName is a linear O(N) scan over all world entities. With a min interval of 10s this can fire frequently.

The transponder entity references should be cached once at `SetHVTAndState` time (they're already fetched there, and their positions are known so just store the IEntity reference):

Same applies in GameInit() — FindEntityByName is called again there per entry.

- `IsHVTAlive` is called from UpdateHVTPositions, SyncTransponderPositions, OnLifeStateChangedWrapper, FindHVTInRange, and CountHVTsInRange. Each call does:

```
SCR_CharacterDamageManagerComponent.Cast(hvtEntity.FindComponent(SCR_CharacterDamageManagerComponent))
```

Component lookups aren't free. These could be cached into m_mEntryToHVT's companion map (e.g. map<IEntity, SCR_CharacterDamageManagerComponent>) at RegisterHVTEntity time, since the component doesn't change.

- The server never clears the event mask in the EOnFrame, even though the clients do. Is this required?

- `FindAndRegisterPlayerHVT` iterates all connected players and calls `GetPrefabData().GetPrefabName()` on each for every unregistered HVT entry. For a full 100-player server with multiple HVT entries this runs every 60s. It's acceptable, but worth flagging if the HVT entry count grows.

TristGit:
You were right about being concerned mainly for what could be easily done withO(1) caching to avoid `FindEntityByName` & `FindComponent` performance hits that're multiplicative of HVT Count + timeBetweenPings. 

- Added reverse lookup map/caching; `m_mEntryToTransponder` - cached by SetHVTAndState 
- Added companion cache map (`map<IEntity, SCR_CharacterDamageManagerComponent>`) to `EntryToHVT` - cached by RegisterHVTEntity and read by IsHVTAlive. Reregistration clears old entities and refires FindAndRegisterPlayerHVT


- EventMask, **_Not Changed_**: The client ends its replication of FixedFrame after `GameInit` after safestart, which is the initial creation of markers while the server continues the loop. A result of replication confusing me, but intent was to prevent clients from also running poll based events after start. lemme know if thats overcomplex or okay.

- `FindAndRegisterPlayerHVT` poll, **_Not Changed_**: I dont know of any / no experience in script invokers for a player reassuming/rejoining into their CRF slot entity to avoid a poll. It hasnt caused issue but willing to learn implementation
But I DID change the polling/reregistering to 120s instead of 60s on each `FindAndRegisterPlayerHVT`. 


Additions:

- Added a Permadeath code and toggle (def: true).
  Essentially adds functionality of dead HVT's to be marked as such, now if there are respawns in HVT mode missions for any reason, the HVT now essentially is just another soldier with no markers/pings/etc. Half failsafe half designer choice. If disabled the HVTs are constantly valid for polling

- Added chat command+func to reset dead HVTs (allows re-registering, for if an HVT dies like stupidly)
Stole from the Prophunt component as I thought it useful

_I also made the obvious addition to `CRF_PlayerRplToOwnerManager`
Inform if this kind of things unneeded_

  `/hvt_reset_deaths`  — `Clears all permadeath-locked PLAYER HVT entries, allowing the now ~120s register poll to take effect if theyre respawned`


I have tested it all in dedicated. 'Works on my machine'. Feel free to test yourself.

PR #1422, opened 2026-05-08
Make CVON killable

tanaaKa:
Enables the killing of cvon which defaults the game back to in-game von systems. Verified all working to include spectators hearing alive players.

### DIFF
--- a/Scripts/Game/Systems/UI/Menus/Spectator/CRF_SpectatorMenu.c
+++ b/Scripts/Game/Systems/UI/Menus/Spectator/CRF_SpectatorMenu.c
@@ -410,7 +410,41 @@ class CRF_SpectatorMenu: ChimeraMenuBase
 		
 		// Handle spectator camera
 		UpdateSpectatorCamera(tDelta);
-		
+
+		// When CVON is disabled, teleport the spectator entity to follow the camera so that
+		// the entity's SCR_VoNComponent picks up direct voice from nearby alive players.
+		// Uses the same pattern as SCR_Global.TeleportPlayer:
+		//   1. GetWorldTransform (world-space, not local) from the camera
+		//   2. BaseGameEntity.Teleport — updates both physics body and scene-graph, and replicates
+		//      the position to the server for the player-owned character (unlike SetTransform alone)
+		//   3. Zero velocities to prevent any residual physics drift
+		if (!CVON_VONGameModeComponent.GetInstance())
+		{
+			IEntity specEntity = SCR_PlayerController.GetLocalMainEntity();
+			if (specEntity && CRF_EntityHelper.IsSpectator(specEntity))
+			{
+				CameraBase camera = GetGame().GetCameraManager().CurrentCamera();
+				if (camera)
+				{
+					vector mat[4];
+					camera.GetWorldTransform(mat);
+
+					BaseGameEntity bgEntity = BaseGameEntity.Cast(specEntity);
+					if (bgEntity)
+						bgEntity.Teleport(mat);
+					else
+						specEntity.SetWorldTransform(mat);
+
+					Physics phys = specEntity.GetPhysics();
+					if (phys)
+					{
+						phys.SetVelocity(vector.Zero);
+						phys.SetAngularVelocity(vector.Zero);
+					}
+				}
+			}
+		}
+
 		// Update follow-mode HUD overlay
 		m_EntityInfoDisplay.UpdateEntityInfoDisplay(m_eSpecEntity);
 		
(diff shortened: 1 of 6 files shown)


Scripts/Game/Systems/UI/Menus/Spectator/CRF_SpectatorMenu.c
```diff
@@ -410,7 +410,41 @@ class CRF_SpectatorMenu: ChimeraMenuBase
...
+					Physics phys = specEntity.GetPhysics();
+					if (phys)
+					{
+						phys.SetVelocity(vector.Zero);
+						phys.SetAngularVelocity(vector.Zero);
+					}
+				}
+			}
+		}
+
```
Njpatman, line 441: Uneeded, use AttatchSpectatorToCamera in the CRF_SpectatorCamera class.

We simply attach them to a virtual bone on the camera, so we don't need to update anything in real time.

Worlds/Arland/CRF_LobbyTestWorld_Layers/AAINT.layer
```diff
@@ -37,12 +37,14 @@ CRF_Gamemode CRF_Lobby : "{6A996BBFCEB37E78}Prefabs/Systems/!Lobby/CRF_Lobby.et"
...
  m_bRespawnEnabled 0
  m_bWaveRespawn 0
+ m_bRallyPointsEnabled 0
  m_iTimeToRespawn 60
  m_iRespawnCutoffMinutes 0
  m_iTimeLimitMinutes 45
  m_bLockUnusedSlots 1
  m_bUseSafestartTimeLimit 0
  m_iSafestartTimeLimit 15
+ m_bUseCVON 0
```
Njpatman, line 47: Revert LobbyTestWorld's CVON attribute back to its default state (true)

Njpatman (changes requested):

Scripts/Game/Systems/UI/Menus/Spectator/CRF_SpectatorMenu.c
tanaaKa: I tried that but it doesn't work.

Njpatman: It should? (at least in my testing a couple of months ago) How did you call it, on the client or on the authority?

tanaaKa: In `CRF_PlayerCameraManager`:`InitilizeSpecCamera()`:Line64

```
if (!CVON_VONGameModeComponent.GetInstance())
    CRF_SpectatorCamera.Cast(m_eCamera).AttatchSpectatorToCamera(m_eCamera);
```

tanaaKa: The camera entity has no rplcomponent and the initial/spectator entity isnt actually attaching/moving with the camera.

Njpatman: AttatchSpectatorToCamera seems to work fine in the initial lobby screen, but doesn't work in the regular spectator menu (from my testing).

Njpatman: image

It's def following the camera around in my dedicated server tools (I commented out the spectator menu code, though it won't work in the lobby/briefing screen anyway)

Njpatman: I also don't know if all clients need to see where the spectator is; the local client may be able to get by with just their local entity being near the other clients to hear their VON.

tanaaKa: Despite that, it doesn't go near alive players on the server, so no spectator ever hears an alive player.

AttachEntity(playerEntity) sets playerEntity's transform to be relative to the camera in the scene graph and the camera doesn't exist on the server. VON proximity checks run server-side against world positions, so the server's computed world position for the spectator is effectively vector.Zero from the deleted attachment parent. That results in it never within range of any alive player.

tanaaKa: I'm all for a more elegant solution though. I just don't see one just yet.

Njpatman: Yeah, that's fair, I'd move the per-frame update to the spectator camera/character though. I don't like the position updates being contingent on the spectator menu being active.

PR #1520, opened 2026-07-08
CRF-TaskCreator.c 

TristGit:
image

## TaskCreator Rough Draft

Idea behind this component is to be a bit modular, with scripters essentially adding their own TaskHandlers (ex. CRF_TaskHandler_CollectIntel.c), insuring the class is added next to the other handlers in the core component, and letting MM's actually continuously work from those changes instead of one-off scripting being scattered throughout.

At the moment is UserAction-centric. Spawns objects like prior gamemode .c's do, and all MM's should have to do is basically choose the behavior of said object while dynamically being able to use any prefab that is setup with the components that are somewhat easy to make.

## CRF Task Creator Component

Generic objective task system for CRF missions. Attach to any gamemode entity to spawn, track, and notify on objective completion per faction.

### Logic Brief
`CRF_TaskCreatorComponent.c`
Core component; spawns prefabs at named anchors, tracks completion, fires win conditions
`CRF_TaskCreatorObjectComponent.c` 
Lightweight component on each spawned task prefab; holds replicated task index/type/state
`CRF_TaskCreatorAction.c` —> `ScriptedUserAction` on task prefabs (not the empty object); delegates all logic to the handler
`CRF_BaseTaskHandler.c` — (Desired objecLogic) —> ex. `CRF_TaskHandler_CollectIntel.c` (Mimics the RUSH bomb planting logic, all within the taskhandler.
The BASE inherited folder to scaffold/override both non-gamemode logic and useraction behavior for task 'types'

`CRF_TaskCreatorPreviewComponent.c`
 Workbench-only; shows task mesh on anchor entities at edit time
`CRF_Task_Empty.et` — pre-configured anchor prefab with preview component attached

### Notes
- Handlers are the only type-aware code; adding a new task type requires nothing in changing the core components OTHER than adding the classes/attribute to the core components OR abnormal logic (stuff I havent accounted for yet)
- Mainly everything so far assumes that it is SCRIPTEDUSERACTION-based, tldr executed via useractions
- Very rough and crude atm
- Only ONE prefab is setup but it should be easy to add more.

- This ALSO makes the RUSH sound playing RPC's/PlayerToAuthority addressable by anything. Did this to let the PlantBomb Taskhandler mimic the RUSH MCOM destruction.

I have added an example setup to the official lobbytest.


Scripts/Game/Systems/Core/Managers/PlayerController/Replication/CRF_PlayerRplToAuthorityManager.c
```diff
@@ -2730,6 +2760,68 @@ class CRF_PlayerRplToAuthorityManager : ScriptComponent
...
 	}
+
+	//------------------------------------------------------------------------------------------------
+	[RplRpc(RplChannel.Reliable, RplRcver.Server)]
+	protected void RpcAsk_RequestObjectiveTaskComplete(int taskIndex, int completingSide)
+	{
+		// Telemetry: int x2
+		LogTelemetry("RpcAsk_RequestObjectiveTaskComplete", CRF_BandwidthTelemetryManager.EstimateSize_Int() * 2);
+
+		CRF_TaskCreatorComponent taskComp = CRF_TaskCreatorComponent.Cast(GetGame().GetGameMode().FindComponent(CRF_TaskCreatorComponent));
```
tanaaKa, line 2771: Should the caller, task state, or whether the action is actually allowed for that player be validated here first?

Scripts/Game/Systems/Components/TaskCreator/Handlers/CRF_TaskHandler_PlantDefuseBomb.c
```diff
@@ -0,0 +1,287 @@
...
+[Attribute("{4BE47BA2B7E3877E}Prefabs/Systems/Fire/Wrapper_Fire_Large_Damage.et", UIWidgets.ResourceNamePicker, "Fire/smoke prefab spawned 385 ms after detonation.", params: "et", category: "Plant/Defuse - Explosion")]
+ResourceName m_sFirePrefab;
+
+//=========================================================================
+// RUNTIME (not serialised)
+//=========================================================================
+
+// Cached world position of the bomb, set just before any entity deletion
+// so deferred effects (e.g. SpawnDelayedExplosionEffects) have a valid pos.
+protected vector m_vExplosionPos;
```
tanaaKa: This is stored on this class instance and shared across all related tasks. If multiple bombs detonate or callbacks overlap, the delayed explosion/fire effects can be spawned at the wrong location because that field gets overwritten by the most recent explosion, right?

tanaaKa (changes requested):

Scripts/Game/Systems/Core/Managers/PlayerController/Replication/CRF_PlayerRplToAuthorityManager.c
TristGit: Its validated within the taskhandler, which carries the faction check into ScriptedUserAction for the object. Meaning its surface level clientside basically. There isnt any validation within the RPCs, No bueno?

tanaaKa: That's fine then

Scripts/Game/Systems/Components/TaskCreator/Handlers/CRF_TaskHandler_PlantDefuseBomb.c
TristGit: Def could cause locality issues regarding the shared vector and cause duplication, addressing

TristGit: > This is stored on this class instance and shared across all related tasks. If multiple bombs detonate or callbacks overlap, the delayed explosion/fire effects can be spawned at the wrong location because that field gets overwritten by the most recent explosion, right?

https://github.com/CoalitionArma/Coalition-Reforger-Framework/commit/51439b1198f1d3b3d26150f734391048bcfa5cc2

PR #1618, opened 2026-08-15
M224 no burn

Hekiao:
M224 No Burn

This Pull Request will:

Add custom M224 Mortar with the white phosphorus burn effects removed from the WP smoke.


tanaaKa:
Isnt the WP burn effect coming from the shell and not the mortar?

